Issue #151, opened 2014-10-20
Debian package


boyska:
This has been faced many times... but now I have a patch for it. If there is interest in the feature, I will propose a clean pull request. You can already see it at https://github.com/AvANa-BBS/Tomb/commit/692723cc0f90221b44a3bc588af4ab6a9ba32f4a

PR #152, opened 2014-10-20
Debian packaging

boyska:
( refs #151 ) This will make "debuild" just work, without adding any complexity to Tomb code. I want to make clear that this is by no means perfect; more specifically:
- I am not sure that every dependency is included (probably not)
- I may be missing some file to install
- It would be very good to also build tomb-extras and other interesting packages, but I still have not figured how to do it. But this is a good start, I hope


boyska:
sorry for the soooo long reply.
all in one commit can be ok, but no way for `extras/debian` to work. The `debian` directory MUST be in the root, or it's not a debian package.

jaromil:
I know, but it can be made a Makefile target that links it or copies it in the root with `make debian`.

I prefer that, I don't want any distribution to pollute the root source directory with their files.

boyska:
>I know, but it can be made a Makefile target that links it or copies it in the root with `make debian`.

which is not debian packaging, unfortunately.

>I prefer that, I don't want any distribution to pollute the root source
>directory with their files.

in that case, the typical solution is to have a "clean" tomb repository
and a separate non-native debian package.
Unfortunately I have no experience with non-native debian packages, so
if "clean root directory" is a requirement for you, I'll just drop the
attempt to include debian packaging in the "official" Tomb repository
and keep it in my fork.

jaromil:
please keep your branch then , we can link it from the docs. Arch already does that.

post a final link to this issue when  ready.

in the future, if we get out of sync, updating upstream by merging your branch is trivial.

I may maintain something similar for Devuan in the future.

boyska:
ok so for now I'll keep a separate branch.
My long-term plan (that is: wishlist) is to have a non-native debian package

>I may maintain something similar for Devuan in the future.

while I commonly use the world "debian", I refer to the packaging
toolchain, not the debian repositories. So you can use the package in
ubuntu, devuan, and probably any other debian-based distribution which
satisfies the dependencies

jaromil:
Sounds great. Is the authoritative branch for your .deb package the one of this pull-req?

jaromil:
I assume this is the branch then, same as pull req https://github.com/AvANa-BBS/Tomb

PR #189, opened 2015-02-16
Detect plain swaps on encrypted partition (util-linux >= 2.22) or advise user to use -f (fixes #163)

### LINKED ISSUE #163: Tomb refuses to open a tomb without -f flag with implicitly encrypted swap
My system resides on an encrypted partition, within which resides LVM, within which reside all my volumes: root, home and swap. Thus, my swap is encrypted. But Tomb doesn't detect that, because it can only detect explicitly encrypted swap.

Can Tomb conduct a deeper analysis of swap partitions? Or would it be too much of an effort, and users of such setups will have to use -f option?

hellekin:
Please see the updated dual behavior.

### DIFF
--- a/tomb
+++ b/tomb
@@ -306,10 +306,44 @@ _ensure_safe_swap() {
     [[ -z "$swaps" ]] && return 0 # No swap partition is active
 
     _message "An active swap partition is detected..."
-    for s in $=swaps; do
-        { _is_encrypted_block $s } \
-	    && { r=2 } || { r=1; break }
-    done
+
+    # Issue #163
+    # lsblk --inverse appeared in util-linux 2.22
+    # but --version is not consistent...
+    local bug_163=0
+    lsblk --help | grep -q '\-\-inverse'
+    if [[ $? -eq 0 ]]; then
+	for s in $=swaps; do
+            { _is_encrypted_block $s } \
+		&& { r=2 } || { r=1; break }
+	done
+    else
+	# Use legacy code that does not detect plain swaps on
+	# encrypted volumes.  On such systems -f must be used.
+	bug_163=1
+	for s in $=swaps; do
+	    bone=$(sudo file $s)
+	    if [[ "$bone" =~ "swap file" ]]; then
+		# It's a regular (unencrypted) swap file
+		r=1
+		break
+
+	    elif [[ "$bone" =~ "symbolic link" ]]; then
+		# Might link to a block
+		r=1
+		[[ "/dev/mapper" == "${s%/*}" ]] || { break }
+		is_crypt=$(sudo dmsetup status "$s" | awk '/crypt/ {print $3}')
+		[[ $is_crypt == "crypt" ]] && { r=2 }
+
+	    elif [[ "$bone" =~ "block special" ]]; then
+		# It's a block
+		r=1
+		is_crypt=`sudo dmsetup status "$s" | awk '/crypt/ {print $3}'`
+		[[ $is_crypt == "crypt" ]] && { r=2 } || { break }
+
+	    fi
+	done
+    fi
 
     if [[ $r -eq 2 ]]; then
         _success "All your swaps are belong to crypt.  Good."
@@ -318,6 +352,10 @@ _ensure_safe_swap() {
         _warning "You can deactivate all swap partitions using the command:"
         _warning " swapoff -a"
         _warning "But if you want to proceed like this, use the -f (force) flag."
+	if [[ $bug_163 -eq 1 ]]; then
+	    _warning "[#163] I cannot detect plain swaps on an encrypted volume."
+	    _warning "[#163] Use -f, or upgrade util-linux to 2.22+."
+	fi
     fi
     return $r
 
--- a/tomb
+++ b/tomb
@@ -2772,7 +2772,7 @@ EOF
 
 # {{{ Run
 
-main "$@" || exit $?   # Prevent `tomb source tomb` from exiting
+main "$@" || exit $?   # Prevent `source tomb source` from exiting
 
 # }}}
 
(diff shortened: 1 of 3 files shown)


hellekin:
This closes #163.

jaromil:
It seems there are conflicts, perhaps easy to solve as they are related to recent small changes. Can you solve them?

Also @boyska can you please re-test this on your environment if the flag is now fixed?

boyska:
@jaromil seems to work me. I have an encrypted lvm, with a swap partition inside, active. If I use my system-wide tomb1.4, it will complain about it. If I check out the pull request, and run `./tomb dig -s 10 test` it works. So, good for me.

I also created a swapfile in my home, and activated it. Running the same test, both tomb 1.4 and this pull request do not complain. Is this expected? does tomb include a mechanism to find out swapfiles on encrypted partitions? unfortunately I do not have a non-encrypted disk, so I can't test if a swapfile on a non-encrypted disk will make tomb complain.

hellekin:
>  does tomb include a mechanism to find out swapfiles on encrypted partitions?

With this fix, yes.  Should we merge?

boyska:
What I am saying is that the swapfile thing does not work. I cannot understand how it could work reading the code. So I decided to test it. I created a swapfile on `/boot`, I `swapon` it, and tomb does not complain. This applies both to 2.0, to the previous version, and to the current master.  So there's no regression, you could say :) anyway it's a bug

hellekin:
@boyska I'm testing it on a fresh machine without encryption...

```
tomb  .  Commanded to open tomb /tmp/test.tomb
tomb  .  An active swap partition is detected...
tomb [W] This poses a security risk.
tomb [W] You can deactivate all swap partitions using the command:
tomb [W]  swapoff -a
tomb [W] [#163] I may not detect plain swaps on an encrypted volume.
tomb [W] But if you want to proceed like this, use the -f (force) flag.
tomb [E] Operation aborted.
     Tomb command returns 1
```

hellekin:
Ah, I see what you mean: it correctly detects unencrypted __blocks__ on encrypted volumes, but not unencrypted __files__ on encrypted partitions.  So I will have to bring back some `file` code to make the case for files.

boyska:
Yep. Since swapfiles are (I believe and I hope) very rare, even something that consider every file to be non-encrypted would be safe.
Something like

```sh
[[ ! -b $b ]] && {r=1; break}
```

should be enough to be on the safe side

hellekin:
Hum, the code __already does this__.   The function `_is_encrypted_block` is well-named: it returns 0 only if the argument is an encrypted __block__.  No, the issue lies elsewhere: a swapfile is never considered secure because it's not a block.  So the code should verify that it's on an encrypted block device to be considered safe.  Otherwise we leave this to the user's consideration.

hellekin:
So, to be precise: I cannot reproduce the behavior you experienced with the swapfile in /boot.  When I do that, Tomb complains about unsafe swap.

hellekin:
I tested again _por las dudas_...

| n | Type of swap    | Container               | Safety           | Test   |
|---|----------------|-----------------|-----------------|---------|
| 1 | no swap            | Any                | Safe             | OK      |
| 2 | encrypted swap  | Any              | Safe             | OK      |
| 3 | plain swap file   | encrypted  | kind of Safe | FAILS    |
| 4 | plain swap block  | encrypted  | kind of Safe | OK    |
| 5 | plain swap   | unencrypted | Unsafe     | OK       | 

The third test is not supported (i.e. the behavior is not changed from previous versions) but the fourth one is supported (which is a new improvement).  As the default behavior is safe (warn user to use -f and abort), I propose to leave it alone for now, unless someone has a neat way to determine whether the file container is an encrypted block without resorting to many lines. (The fix should be shorter than this comment ;o)

jaromil:
very good overview. I think we can live with this.

tomb
```diff
@@ -278,6 +278,25 @@ _tmp_create() {
...
+# Synopsis: _is_encrypted_block /path/to/block/device
+# Return 0 if it is an encrypted block device
+_is_encrypted_block() {
+    local    b=$1 # Path to a block device
+    local    s="" # lsblk option -s (if available)
+
+    # Issue #163
+    # lsblk --inverse appeared in util-linux 2.22
+    # but --version is not consistent...
+    lsblk --help | grep -q '\-\-inverse'
```
boyska: `grep -Fq -- --inverse` is much more readable!

boyska:
TL;DR: everything was fine, sorry for getting confused.

Ok, I was testing with the `tomb dig` subcommand, but `dig` does not invoke `check_swap`. This could be intentional, I don't know if it is correct (I would `check_swap` in each and every case!), but it's surely not related to this pull request.

Now I am testing with `tomb open` which indeed ensures that swap is not used. Now I see that swapfiles are never considered to be encryped. This is not correct (and wasn't correct even before, so no regression) but is safe.

tomb
hellekin: Done @8aa1c0c6a5d98f19f76ca9737ee7d136a9d13fa5 Thank you @boyska!

hellekin:
Added comments in @8aa1c0c6a5d98f19f76ca9737ee7d136a9d13fa5 for people who feel like supporting the case for plain swapfile over encrypted filesystem.

jaromil:
yes not checking for swap in functions that do not deal with keys is intentional

boyska:
even `lock` doesn't check_swap, which seems a security issue to me.

If the rationale is "who uses gpg must check swap", maybe the best thing is to run `_check_swap` inside of `gpg_decrypt` and other "sensitive" places. Not sure about this code style, though.

jaromil:
Well spotted.

I'd rather call it in the header of outer functions.

We may to go through this and check that is everywhere it needs...

Perhaps easier to just put it everywhere as you mentioned before, I'm now convinced.

So then it would go in the main()

hellekin:
I thought `_check_swap` was already called in `main()`.  Maybe something I forgot to keep from the refactoring.  Or was it removed at some point?

PR #190, opened 2015-02-17
Remove outdated error message

jturner314:
Commit 5dbcabdf2636300f6877e6bab27ef51c7b739a8f removed the `chown` call in `_tmp_create()` but not the associated error message. This commit removes that error message since it is just dead code.


gdrooid:
Thank you!

PR #191, opened 2015-02-18
Remove redundant test for $DISPLAY

jturner314:
This block is already wrapped in the `else` block of the same test for the `$DISPLAY` variable, so this test is redundant.


jaromil:
thanks for carefully reading all the code and checking for such imperfections.

jturner314:
You're welcome. I'm in the process of reviewing the code before I start using it myself, so I may send a few more pull requests as I notice things like this.

PR #193, opened 2015-02-20
FIX errors on kdf iteration count

### LINKED ISSUE #192: KDF iteration time is wrong
The `--kdf` option allows the user to enforce kdf. Its argument is said (not very clearly, actually) to be the number of seconds that it will take to do stretching on that box.

Since the `getiter` program takes an input the microseconds, the number of seconds should be multiplied by one million. Instead, it is multiplied by 10k. see https://github.com/dyne/Tomb/blob/master/tomb#L1141
It is even documented: https://github.com/dyne/Tomb/blob/master/doc/tomb.1#L198

I propose to:
* fix the multiplier
* make a clearer documentation

and I am willing to do it myself, if you agree.

boyska:
Both the code and the documentation were incorrect about --kdf.

The code was using an iteration time much inferior to the one that the user could be expecting. The output messages were even more confusing, calling "count" what was actually "time".

The same error was present in the man page.

This commit tries to fix both of them.

This will close #192

### DIFF
--- a/doc/tomb.1
+++ b/doc/tomb.1
@@ -193,10 +193,12 @@ what you are doing if you force an operation.
 When digging or resizing a tomb, this option must be used to specify
 the \fIsize\fR of the new file to be created. Units are megabytes (MiB).
 .B
-.IP "--kdf \fI<iterations>\fR"
+.IP "--kdf \fI<itertime>\fR"
 Activate the KDF feature against dictionary attacks when creating a
-key: forces a delay of \fI<iterations>\fR (integer multiplied by 10k)
-every time this key is used.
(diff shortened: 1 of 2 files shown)


jaromil:
as mentioned in #192 this likely breaks previously created tombs.

it was merged too quickly and it must be now reverted.

we will not fix this now and eventually we will tackle the problem altogether with a switch to gnupg's implementation of -s2k which is likely to be compatible with current code.

boyska:
wait. You're just wrong, or so I think. I only edited the `gen_key` function, so it doesn't change a thing on already generated keys.

PR #194, opened 2015-03-01
Two minor one-line fixes

jturner314:
The `exhumedkey` variable is initialized but never used, so this PR removes it. The comment removed by this PR is inaccurate. (The following line does not print the password.)


jaromil:
thanks

PR #199, opened 2015-04-20
Fix typo in French translation

arthaud:


extras/po/fr.po
```diff
@@ -882,7 +882,7 @@ msgstr ""
 #: tomb:Open:mount_tomb:1830
 msgid "Last visit by ::1 user::(::2 tomb build::) from ::3 tty:: on ::4 host::"
 msgstr ""
-"Dernière visite par ::1 user::(::2 tomb build) depuis ::3 tty:: sur ::4 "
+"Dernière visite par ::1 user::(::2 tomb build::) depuis ::3 tty:: sur ::4 "
 "host::"
```
hellekin: Shouldn't this be on the previous line?

arthaud: Why? The "msgid" line is OK.

Narrat: I suppose he asked if this second line of msgstr shouldn't be actually the first one.

Why is it splitted anyway?

jaromil:
@gdrooid will kindly answer when he has time, he is maintaining the translations.

extras/po/fr.po
arthaud: Oh yes, you are right.

As you can see, I just updated the second line, it was already split. I guess it's because of the 80 characters limit.

gdrooid:
This is alright, but I'll close this PR and merge #200.

extras/po/fr.po
gdrooid: I can't remember if it's POEditor or `msgmerge`, but one on them is splitting the lines every time I update translations, so it better stays like that.

PR #208, opened 2015-07-08
Man-Page: Small fixes

Narrat:
Mostly intended to fix the URLs at the end of the man-page.
The original page for dm-crypt pointed to the google-code page of cryptsetup and the google-code page itself redirects the user to gitlab, where the project is now hosted.
So replaced the URLs with their gitlab counterpart.
I inserted the extra .br because of the long URL. Without it those get fragmented even on larger terminal windows. But dunno if's that wise?

Made a note about the mount behaviour which got enhanced by ca012e8

Additionally some smaller typos corrected


jaromil:
very nice tidying up, thanks!

PR #215, opened 2015-09-02
Added tomb-qt-tray

mgian:
tomb-qt-tray is an equivalent of gtk-tray, only it is based on the Qt library


jaromil:
excellent! can you please add a license header on each source file?
can be copied from other sources, but add your name or nickname and year

mgian:
Added to all source files. (well, except the .ts file(s) since with the copyright notice, the linguist-qt* software don't parse them)

jaromil:
Great, thanks! I'm testing the build and noticed it requires QT-5.4.

There should be a README.txt in extras/qt-tray to indicate requirements, but ok I can add that myself after merging the pull request.

mgian:
Thanks. Just a quick note: my correct name is Gianluca, not Michele ;-)

jaromil:
Oops :^) sry

Issue #229, opened 2016-10-09
Proposed change for when encrypted swap is found


mandeep:
Currently when tomb finds an encrypted swap partition, the following message is sent to stdout:
```
"All your swaps are belong to crypt. Good."
```

I propose changing this to something more informative along the lines of:
```
Encrypted swap partition found. Swap partition is safe.
```

Thoughts?

hellekin:
> Thoughts?
> 

Doesn't it lack personality?

mandeep:
> Doesn't it lack personality?

I agree it's a bit dry, do you have a statement in mind that is both whimsical and informative?

Maybe something like
```
Encrypted swap partition found. Your swap is safe from the undertaker.
```
or
```
The undertaker found an encrypted swap partition. Good.
```

jaromil:
Perhaps the second. Hassling to have to change all translations.. actually I'm curious how this joke has been translated...

mandeep:
Can the original message even be translated to other languages? Seems like something that only exists in English.

jaromil:
Its all in `extras/translations`

The italian one kept literal: "Tutti i tuoi swap appartengono a crypt. Bene."
which indeed doesn't sounds informative nor as funny as in English.

hellekin:
Most jokes or poems are not translatable.  Colloquial expressions that
are not translatable literally are usually translated to another,
similar expression that conveys the same meaning in the culture
associated with that language.

"[All your base are belong to
us](https://en.wikipedia.org/wiki/All_your_base_are_belong_to_us)" has a
specific history and context that conveys a certain humor and
**culture** but it would be a shame to abandon it for that reason.  If
you really want to "do it", why not take the translations used by
others, like in Wikipedia, that translate, e.g., [in
French](https://fr.wikipedia.org/wiki/All_your_base_are_belong_to_us):

> Toutes votre base sont appartiennent à nous.

Or [in
Italian](https://it.wikipedia.org/wiki/All_your_base_are_belong_to_us):

> Tutte le vostre base sono appartengono a noi.

I would vote to use **these** instead of the more classical,
_informative_, **utilitarian** versions that we have now, that hardly
convey any personality.  If people have issues understanding what that
means, at least they will learn something, and enjoy some **human**
experience.

mandeep:
[...]

PR #230, opened 2016-10-11
Changed message when encrypted swap found to something more informative

mandeep:
In relation to #229, I found the current message to be a bit ambiguous. When I first started using Tomb, I wasn't sure if the success message was a good one or a bad one. Hopefully a clearer message will let new users understand that their swap is encrypted.

### DIFF
--- a/tomb
+++ b/tomb
@@ -325,7 +325,7 @@ _ensure_safe_swap() {
     done
 
     if [[ $r -eq 2 ]]; then
-        _success "All your swaps are belong to crypt. Good."
+        _success "The undertaker found that all swap partitions are encrypted. Good."
     else
         _warning "This poses a security risk."
         _warning "You can deactivate all swap partitions using the command:"


mandeep:
How do we determine how many swap partitions are actually found? Is having more than one swap partition common?

gdrooid:
I don't know if it's common to have more than one swap partition, but [we're checking that all of them are encrypted](https://github.com/mandeep/Tomb/blob/ea58e5094e7de6d1945aabd8fff3f9c391539139/tomb#L313-L325).

mandeep:
Rebased :smiley:

hellekin:
Sad day for culture.

gdrooid:
😅 I'll push the change to POEditor soon.

mandeep:
@hellekin I think Tomb will be okay with this minor change :).

Thanks everyone for letting me work on this.

PR #243, opened 2017-02-03
Use of lsof to fix slam for specific mountpoint

### LINKED ISSUE #220: fuser vs lsof (slam fails, can't reproduce yet)
I am running arch linux with the AUR tomb-git + tomb-kdf-git package.
Everytime I enter "slam all" or "slam all -f" while having a program accessing a file in my active tomb, I get "target is busy" and try fuser(1).

I hope, that this is the right place for this issue.
Good evening, gentlemen

Narrat:
Apparantly fuser didn't report back, if the tomb was mounted in a subdir of /run (whereas /run itself is often a tmpfs mount).
With no list of process ids those couldn't be killed, so slamming the tomb failed.
lsof is capable to report back the sought information.

Fixes #220

Additionally fixing the debug output, where a hardcoded mountpoint was used

_______________
Well, that would be my proposal of replacing fuser.
Testing worked for me :D
Any comments, regarding flaws and oversoughts?


jaromil:
Seems great, will test ASAP. What comes to mind is perhaps to make the `slam` command "optional" when `lsof` is installed, just like with `bury/exhume` commands depending from `steghide`.

Narrat:
Makes sense. Will add it

Narrat:
I made it an optional feature.
But I decided to not error out, if lsof is not found. Instead it switches back to try a normal close, which could work in the case, no program is accessing something inside the tomb

jaromil:
Perfect! thanks! sooner than later we shall add a test unit for the slam command as well.

Narrat:
Gosh.. embarrassing :D
I noticed one small mistake, I added while doing the final cleanup (Without, of course, rechecking).
For whatever reasons I first used 
>lsof -h  1>/dev/null 2>/dev/null || LSOF=0
-h because it was faster.
And I just added command -v in front, like the rest. Now I notice this causes LSOF to be set everytime. So another PR incoming *cough* Sorry for that

Issue #179, opened 2014-12-08
Do not prompt for Tomb password if secret key material doesn't need one


alphazo:
There is one little last issue with the GnuPG setup. When I open the Tomb container with a GnuPG encrypted keyfile I first get prompted for a Tomb password for which I can type anything since it no longer applies and then I get the GnuPG password prompt.
Would be nice if Tomb first tried to use the GnuPG _unprotected_ secret instead of prompting for a password?
## Generating a 100MB empty tomb

```
# tomb dig -s 100 test.tomb
 tomb  .  Commanded to dig tomb test.tomb
 tomb (*) Creating a new tomb in ./test.tomb
 tomb  .  Generating test.tomb of 100MiB
 100+0 records in
 100+0 records out
 104857600 bytes (105 MB) copied, 8,55695 s, 12,3 MB/s
 -rw------- 1 alpha users 100M  7 déc.  12:20 ./test.tomb
 tomb (*) Done digging test
 tomb  .  Your tomb is not yet ready, you need to forge a key and lock it:
 tomb  .  tomb forge ./test.tomb.key
 tomb  .  tomb lock ./test.tomb -k ./test.tomb.key
```
## Creating the encryption key and using a dummy temporary password to protect it

Password in that case is "weak".

```
 tomb forge test.tomb.key
 tomb  .  Commanded to forge key test.tomb.key
 tomb [W] An active swap partition is detected, this poses security risks.
 tomb (*) All your swaps are belong to crypt.  Good.
 tomb  .  Commanded to forge key test.tomb.key with cipher algorithm AES256
 tomb  .  This operation takes time, keep using this computer on other tasks,
 tomb  .  once done you will be asked to choose a password for your tomb.
 tomb  .  To make it faster you can move the mouse around.
 tomb  .  If you are on a server, you can use an Entropy Generation Daemon.
 256+0 records in
 256+0 records out
 256 bytes (256 B) copied, 111,361 s, 0,0 kB/s
 tomb (*) Choose the  password of your key: test.tomb.key
 tomb  .  (You can also change it later using 'tomb passwd'.)
 [GNUPG:] PROGRESS need_entropy X 8 16
 [GNUPG:] PROGRESS need_entropy X 16 16
 [GNUPG:] NEED_PASSPHRASE_SYM 9 3 2
 [GNUPG:] BEGIN_ENCRYPTION 2 9
 [GNUPG:] END_ENCRYPTION
 tomb  .  Key is valid.
[...]

PR #244, opened 2017-02-04
GnuPG Key Support

roddhjav:
This pull request tries to answer the issue #179 adding full support of asymmetric GnuPG key to encrypt a tomb key. This branch was original based on the `gnupg-delegation` branch.

# Description
This branch adds the following *common* features to tomb:
* Forge, lock, open a tomb key encrypted with a GPG key.
* Change the GPG key used to encrypt the tomb.
* Bury, exhume the key tomb with a steganography password (while the tomb key stays encrypted with the GPG key)
* Change the tomb key for a tomb (setkey)

Moreover adding the support to the de-facto GPG encryption standard give to tomb the following *new* feature:
* Sharing support: A tomb key can be encrypted for more than one GPG key.
* OpenPGP Smart Card support.

## The new options: `-g`, `-r`
These features are activated using the option `-g/--gpgkey `.  The option `-r` (for recipient like in GPG itself) is used to provide the gpg recipient.  `-g` and `-r` options can be used together in the following commands:
* `tomb forge`: Because only the public key is needed, GPG will never ask you for a password. However tomb checks if you have the corresponding private key in your GPG database.
* `tomb lock`
* `tomb open`: The `-r` option is not mandatory.
* `tomb passwd`: It is actually not possible to *update* a key from a passphrase encrypted to a GPG key encrypted, but it makes sense, I plan to do it. However, it will never be possible to *downgrade* a key from  GPG key encrypted to passphrase encrypted.
* `tomb setkey`: Only the change of GPG encrypted tomb key has been tested. I'm not sure if it makes sense to do otherwise.
*  `tomb bury`: see [Steganography](#Steganography)
* `tomb exhume`: see [Steganography](#Steganography)

## Sharing support
A tomb key can be encrypted with more than one recipient. Therefore, a tomb can be shared between different user. The multiple recipients are given using the `-r` option and must be separated by a coma `,`. However, this is a very sensitive action, and the user needs to trust the GPG public key it is going to share its tomb. This is why **this feature needs to be explicitly activated using in more the flag `--shared`**. The `--shared` option can be used in `tomb forge`, `tomb setkey` and `tomb passwd`.

## OpenPGP Smart Cards
Are supported, you need to set it up in your GPG database and it should works out of the box. It provides an extra level of security and allow a PIN code to be used for unlocking the GPG key and decrypt the tomb key. After 3 false attempts the smart card locks itself down.

## Steganography
The tomb policy is to use the same password to encrypt the key and to bury it. However, steganography cannot be done with GPG key. Therefore, we check if the user can decrypt the tomb with its GPG key and we ask for a steganography password. Having different method is a technical requirement and should enhance security.

# Documentation
See the man page with the description of these changes: `doc/tomb.1`

# Examples:

**Create and open a tomb**
```sh
$ tomb dig test.tomb -s 20

$ tomb forge test.tomb.key -gr F0E573B8289439CD
tomb  .  An active swap partition is detected...
[sudo] password for <user>: 
tomb (*) The undertaker found that all swap partitions are encrypted. Good.
tomb  .  Commanded to forge key test.tomb.key with cipher algorithm AES256
tomb [W] This operation takes time, keep using this computer on other tasks,
tomb [W] once done you will be asked to choose a password for your tomb.
tomb [W] To make it faster you can move the mouse around.
tomb [W] If you are on a server, you can use an Entropy Generation Daemon.
256 blocks (0Mb) written.
256+0 records in
256+0 records out
tomb (*) Using the GnuPG key F0E573B8289439CD to encrypt the key: test.tomb.key
tomb  .  (You can also change it later using 'tomb passwd'.)
tomb  .  Key is valid.
tomb  .  Done forging test.tomb.key
tomb (*) Your key is ready:

$ tomb lock  test.tomb -k test.tomb.key -gr F0E573B8289439CD
tomb  .  Commanded to lock tomb test.tomb
tomb  .  Key encrypted with a GnuPG Key
tomb  .  Key is valid.
tomb  .  Locking using cipher: aes-xts-plain64:sha256
tomb (*) Locking test.tomb with test.tomb.key
tomb  .  Formatting Luks mapped device.
tomb  .  Formatting your Tomb with Ext3/Ext4 filesystem.
tomb  .  Done locking test using Luks dm-crypt aes-xts-plain64:sha256
tomb (*) Your tomb is ready in test.tomb and secured with key test.tomb.key

$ tomb open test.tomb -k test.tomb.key -g
tomb  .  Commanded to open tomb test.tomb
tomb  .  An active swap partition is detected...
tomb (*) The undertaker found that all swap partitions are encrypted. Good.
tomb  .  Valid tomb file found: test.tomb
tomb  .  Key encrypted with a GnuPG Key
tomb  .  Key is valid.
tomb  .  Mountpoint not specified, using default: /run/media/<user>/test
tomb (*) Opening test on /run/media/<user>/test
tomb  .  This tomb is a valid LUKS encrypted device.
tomb  .  Cipher is "aes" mode "xts-plain64:sha256" hash "sha256"
tomb (*) Success unlocking tomb test
tomb  .  Checking filesystem via /dev/loop3
fsck from util-linux 2.29
test: clean, 11/2048 files, 1362/8192 blocks
tomb (*) Success opening test.tomb on /run/media/<user>/test
```

**Steganography commands**
```sh
$ tomb bury tomb.jpg -k test.tomb.key -gr F0E573B8289439CD
tomb  .  Key encrypted with a GnuPG Key
tomb  .  Key is valid.
tomb (*) Encoding key -----BEGIN PGP MESSAGE-----
-----END PGP MESSAGE-----
 inside image tomb.jpg
tomb  .  Using GnuPG Key ID
embedding standard input in "tomb.jpg"... done
tomb (*) Tomb key encoded successfully into image tomb.jpg

$ tomb exhume tomb.jpg -k exhumed.tomb.key

$ # If you want to open the tomb with a buried key
$ tomb open test.tomb -k tomb.jpg -gr F0E573B8289439CD
```

**Change the gpg key id**
```sh
$ tomb passwd -k /test.tomb.key -gr 3DC836481F44B31E
tomb  .  An active swap partition is detected...
tomb (*) The undertaker found that all swap partitions are encrypted. Good.
tomb  .  Key encrypted with a GnuPG Key
tomb  .  Key is valid.
tomb  .  Commanded to change GnuPG key for tomb key test.tomb.key
tomb (*) Changing GnuPG key for test.tomb.key
tomb  .  Key is valid.
tomb (*) Your GnuPG key was successfully updated
```

**Resize a tomb**
```sh
$ tomb resize -s 20 test.tomb -k test.tomb.key -gr F0E573B8289439CD
```

**Set a new tomb key**
```bash
$ tomb setkey -k new.test.tomb.key test.tomb.key test.tomb -gr F0E573B8289439CD
tomb  .  Commanded to reset key for tomb test.tomb
tomb  .  An active swap partition is detected...
tomb (*) The undertaker found that all swap partitions are encrypted. Good.
tomb  .  Valid tomb file found: test.tomb
tomb  .  Key encrypted with a GnuPG Key
tomb  .  Key is valid.
tomb (*) Changing lock on tomb test
tomb  .  Old key: test.tomb.key
tomb  .  Key encrypted with a GnuPG Key
tomb  .  Key is valid.
tomb  .  New key: new.test.tomb.key
tomb (*) Succesfully changed key for tomb: test.tomb
tomb  .  The new key is: new.test.tomb.key
```

**Share a tomb**
```bash
$ tomb dig  shared.tomb -s 10
$ tomb forge shared.tomb.key -gr F0E573B8289439CD,3DC836481F44B31E --shared
tomb  .  An active swap partition is detected...
tomb (*) The undertaker found that all swap partitions are encrypted. Good.
tomb  .  Commanded to forge key shared.tomb.key with cipher algorithm AES256
tomb [W] This operation takes time, keep using this computer on other tasks,
tomb [W] once done you will be asked to choose a password for your tomb.
tomb [W] To make it faster you can move the mouse around.
tomb [W] If you are on a server, you can use an Entropy Generation Daemon.
256 blocks (0Mb) written.
256+0 records in
256+0 records out
tomb (*) Using GnuPG keys to encrypt and share your key: shared.tomb.key
tomb  .  (You can also change it later using 'tomb passwd'.)
tomb [W] You are going to encrypt a tomb key with 2 recipients.
tomb [W] It is your responsibility to check the fingerprint of these recipients.
tomb [W] The fingerprints are:
tomb [W]     A485 7CD1 76B3 1435 F970 9D25 F0E5 73B8 2894 39CD 
tomb [W]     0B22 35E6 6075 3AB0 475F B3E2 3DC8 3648 1F44 B31E 
tomb  .  Key is valid.
tomb  .  Done forging shared.tomb.key
tomb (*) Your key is ready

$ tomb lock shared.tomb -k shared.tomb.key -gr F0E573B8289439CD
```

# Going Technical
## Test `extras/test`
I added a lot of tests in the test suite. All the tests (new one, classic and regression tests) passes. I also created a GPG database in `extras/test/gnupg` with two (unencrypted) keys for the test suite use. You can see the details of every build on my [Gitlab dev repo](https://gitlab.com/roddhjav/tomb/builds).

## Tomb key encryption/decryption

They are the modifications I did on the encryption/decryption part. The changes are mostly due to:
* Asymmetric key support (eg: `--recipient` instead of `--symmetric`...)
* No passphrase with GPG key.

### Key decryption  `gpg_decrypt`
```zsh
    TOMBSECRET=`print - "$gpgpass" | \
        gpg --batch ${gpgpopt[@]} --no-tty --no-options \
            --status-fd 2 --no-mdc-warning --no-permission-warning \
            --no-secmem-warning 2> $tmpres`
```
**With:** 

|      | **passphrase** | **GPG key** |
|----|----|----|
| `$gpgpass` | `"${tombpass}\n$TOMBSECRET"` | `"$TOMBSECRET"` |
| `"$gpgopt"` | `(--passphrase-fd 0)` | `()` |

### Key Encryption `gen_key`
```zsh
    print $opt - "$gpgpass" \
        | gpg --openpgp --force-mdc --cipher-algo ${algo} --batch \
              --no-options --no-tty ${gpgopt[@]} \
              --status-fd 2 -o - --armor 2> $tmpres
```
**With:**

|      | **passphrase** | **GPG key** |
|----|----|----|
| `$gpgpass` | `"${tombpass}\n$TOMBSECRET"` | `"$TOMBSECRET"` |
| `$gpgopt` | `(--passphrase-fd 0 --symmetric)` | `(--encrypt `_recipients_arg $recipients`)` |
| `$opt` | `'-n'` | `''`

## New functions
* `is_valid_recipients` All the keys ID must be valid (the public keys must be present in the database), moreover, at least one private key must be present
* `_recipients_arg` Print the recipient arg to be used in gpg.
* `_fingerprint` Print the fingerprint of a GPG key

# Feedback
Are welcome! If you do not like a thing in the implementation please tell me, I will be happy to change anything that need to be changed.


parazyd:
Indeed this looks like it will do the trick. I can also test it out with a gpg smartcard this week if needed.

saro:
Just tested with a smartcard and it seems to work correctly. I have only one request. Could be possible to add an additional flag like `-ra` (recipient automatic) to replace the `-r <dummy> -f` case?
Thanks

roddhjav:
Yes, it is possible. I also don't like the `-r <dummy> -f` case. I just don't know which flag could be good.
Moreover, it could make some confusions: using `-r` everywhere to set the recipient but not in `tomb open`. This is against tomb policy on options.

Any idea for a clean solution?

jaromil:
Options can be either global or bound to commands. In case of `-r` it may make more sense to make it bound to (many) commands, but not global. But I'm not sure this would solve the problem, as I did not look into this yet, just trying to clarify the policy.

roddhjav:
`-r` option is bound to (many) command (like `-k`). But `-k` **always** requires an argument. Here when opening a tomb (and only when opening a tomb) we don't need the recipient. Therefore the simple solution would be to say: with `tomb open`, `-r` doesn't require an argument. Tell if if that suit you.

Also, GPG support hidden recipient (`--hidden-recipient` instead of `--recipient` in GPG). I think it could be the default encryption option. I did the change locally, it runs fine.

jaromil:
ACK, then no it doesn't suits: the number of arguments for the same option flag should be consistent. The `-r anything` solution is yukie.
The alternative is perhaps having a global `-g / --gpgkey` option without any argument to switch on the gpg-key-support across all commands. They would then give an error in absence of `-r`, with exception of the open command (which ignores it). This may also facilitate in future the adding of some more --gpgkey specific options.

Regarding switching on `hidden-recipient` by default yes, I think that is a good idea.

roddhjav:
I did the changes you proposed:

For the different options:
* Added `-g/--gpgkey` global option to tell tomb to use GPG key to encrypt a tomb key.
* Option `-r` is now only used to provide the recipient.
* Option `-R` removed, the new recipient can be given by the -r option.

For the hidden recipient:
* Use `--hidden-recipient` instead of `--recipient`.
* Due to the hidden recipient, GPG will try all the available keys.
* User can speed up this process providing the recipient using the -r option.
* Therefore, `tomb open` optionally support the -r option.

saro:
Just tested with two smartcards the latest version of the pull request. It works correctly and finally I don't need to create two different keys manually for my smartcards. The only small annoyance is that when I perform:
`tomb open <tomb> -k <shared_key> -g`
with the second card inserted (just to be clear, the second one I used in `tomb forge` `-r` option) I get a request to insert card1 after the PIN. This appears also if I add `-r <key_card2>` option to the command.

roddhjav:
When you use `tomb open <tomb> -k <shared_key> -g` this is a normal behaviour. Because the GPG ID is not present in the tomb key, therefore GPG tries all the keys in its keyring. This is why I added the option to do `tomb open <tomb> -k <shared_key> -g -r <key_card2>`. This option uses the GPG option `--try-secret-key` only available since GPG 2.1. Do you have GPG >= 2.1? If yes this is not a normal behaviour and I will need to do more tests on this.

saro:
I'm using gpg 2.1.17. What I find particularly strange is the request for a card1 after I insert the correct PIN for card2. Let me know if I can perform any specific test to help you.

roddhjav:
The issue might be how I use the `--try-secret-key` option. It worked fine with the keyring in `extra/tests/gnupg` but in this keyring all the key are unencrypted.

If you want to help me on this, you can to encrypt a file with the -R option:
```cat file | gpg -e -R <key1> -R <key2> > file.gpg ```
And ask him to decrypt it using with:
```cat file.gpg | gpg -d --try-secret-key <key2> > file.gpg ```

**Comment by saro** [comment]
I can still see the request to insert card1 after the PIN.
Checking the `man gpg` I found the following:
```
       --try-secret-key name
              For  hidden  recipients  GPG needs to know the keys to use for trial decryption.  The key set with --default-key is always tried first, but this is often
              not sufficient.  This option allows setting more keys to be used for trial decryption.  Although any valid user-id specification may be used for name  it
              makes  sense to use at least the long keyid to avoid ambiguities.  Note that gpg-agent might pop up a pinentry for a lot keys to do the trial decryption.
              If you want to stop all further trial decryption you may use close-window button instead of the cancel button.
```
Is it possible that this  "trial decryption" is the reason behind the unnecessary card1 request?
In any case, personally doesn't look like a issue for me and I would be happy to see the gpg support land on master even with this small annoyance.

**Comment by roddhjav** [comment]
I had the same issue:
```
cat enc | gpg -d --try-secret-key <key_1> -d
gpg: anonymous recipient; trying secret key <key_1> ...
gpg: anonymous recipient; trying secret key <key_2> ...
gpg: anonymous recipient; trying secret key <key_3> ...
gpg: anonymous recipient; trying secret key <key_1> ...
gpg: okay, we are the anonymous recipient.
...
```
It seems GPG always tries all the private key even thought the first one is the good one.

**Comment by linux-modder** [comment]
@roddhjav any chance you could add support to pull the default-key  from said  gpg database in place of the -r <dummy> -f ( also this would I'd think make it more robust and seamless with the 2.1.x dirmgr and smartcard ( may also fix the pin1 pin2 issue @ssaro was mentioning as it would first check that key ( as it would conceivably be the one that user is hoping to use on that machine ( save a public shared corporate style machine)

**Comment by jaromil** [comment]
About the `--hidden-recipient` + `--try-secret-key` issue, I think GPG developer may be interested to know this and eventually correct it so that not all secret keys are tried when one is explicitly specified. Let me know if you are comfortable with dropping a note on the g10 Code's bug tracker, else I can also do it.

**Comment by roddhjav** [comment]
@linux-modder Do you mean always use the default-key unless the `-r` option provide a recipient for all gpgkey related command? Yes it is possible, I will work on it.

@jaromil I don't know if it is a bug or a feature. I mean maybe I don't use the option correctly.

**Comment by linux-modder** [comment]
Make that the default was my thought and short of that being declared in distro specific database  (or being in that machines secring.gpg yes fallback to the -r  ( like gpg 2.1.x does now)

**Comment by roddhjav** [comment]
@linux-modder Do you know if it is possible to use hidden recipient with the default key.

**Comment by jaromil** [comment]
Dear Alexandre, I have read and tested your `gnupg-key-support` branch and find it a very well thought contribution, better than what I could do with the previous `gnupg-delegation` branch which never really made me proud enough to merge in master to be honest.

If you agree with merging at this stage, I'll proceed accepting this pull-request. Then the corner case you are currently debating can be moved to a specific issue.

**Comment by roddhjav** [comment]
That's a great new!

Yes the PR is "stable enough" to be merged at this stage. The only reaming task would be to update the man page with the last options (new `-g`, `-r` role changed, no more `-R`). However, because I'm not a native English, maybe I'm not the best person to do it ;).

I'll create a issue for the minor changes we are debating and when the time come I'll create a PR.

**Comment by jaromil** [comment]
Great! again thanks and welcome among the Tomb contributors!

**Comment by stevesbrain** [comment]
What a brilliant contribution! Been waiting for something like this for a long time - I use my SmartCard for just about everything :)

**Comment by alphazo** [comment]
I didn't imagine that opening the issue  #179 would have triggered such amount of great code. Thanks again to all the contributors.
I went through the getting started and experienced small issues:
- I had to use -r even when opening the tomb since I have multiple private keys in my keyring. It started by asking me a password for the first private key I own (no smart card on that one) and then after hiting cancel went to the next one and so on until reaching the one that is linked to the smartcard. I thought the key ID was stored somewhere along the key.
- Mounting is done with root rights so my regular user cannot access the mount point (even if changing the mount point during the open operation).

```
tomb open test.tomb -k test.tomb.key -gr CC7F3AC7B85B4EB8
tomb  .  Commanded to open tomb test.tomb
tomb  .  An active swap partition is detected...
tomb (*) The undertaker found that all swap partitions are encrypted. Good.
tomb  .  Valid tomb file found: test.tomb
tomb  .  Key encrypted with a GnuPG Key
tomb  .  Key is valid.
tomb  .  Mountpoint not specified, using default: /run/media/alpha/test
tomb (*) Opening test on /run/media/alpha/test
tomb  .  This tomb is a valid LUKS encrypted device.
tomb  .  Cipher is "aes" mode "xts-plain64:sha256" hash "sha256"
tomb (*) Success unlocking tomb test
tomb  .  Checking filesystem via /dev/loop1
fsck from util-linux 2.29.2
test: clean, 11/4608 files, 1911/18432 blocks
tomb (*) Success opening test.tomb on /run/media/alpha/test
mount_tomb:135: permission denied: /run/media/alpha/test/.uid
mount_tomb:137: permission denied: /run/media/alpha/test/.tty
mount_tomb:140: permission denied: /run/media/alpha/test/.host
mount_tomb:144: permission denied: /run/media/alpha/test/.last
```

roddhjav:
Hi @alphazo, Thank for the feedback.

For your first point, this is because we use the `hidden-recipient` option of GPG. However, as you wrote, this is not a perfect solution in term of usability. I fixed it #250.

For your second point, this is not an issue related to the GPG support but (I think) to 843b7fdfc4c125065d31cc11cff8a994ed680bd4

alphazo:
Merci @roddhjav for the feedback. I have withdrawn my second comment regarding the mounting issue but  you were still able to catch it and point me to the right known issue. I understand that using `hidden-recipient `might make sense in many situations but I like the flexibility offered by your solution too.

Issue #255, opened 2017-04-05
Locally Decrypted GPG over SSH


stevesbrain:
Hi all,

I have a scenario (like I'm sure some of us do) where my tombs themselves live on remote servers, and I would previously have to pipe my key over SSH like:

```bash
ssh my.other.server.net 'cat tomb.key' | tomb open -k - secrets.tomb 
```

Now, when switching to using an existing GPG key, the private key for that lives only on my YubiKey. This works great locally - but how can I go about decrypting this over SSH? The remote system has no access to the private key, and piping like:

```bash
gpg --decrypt tomb.key | tomb open -k - secrets.tomb
```

Does not work, of course. Has anyone else had any success in this same scenario, or have any suggestions? Thanks!

jaromil:
Hi! When you write "when switching to using an existing GPG key", do you mean a key that is asymmetrically encrypted using the new `-g` switch, recently implemented in our development version here by @roddhjav ?

roddhjav:
One of the first commit I did with the GPG support introduced a bug preventing to support of gpg encrypted tomb key loaded from stdin. This is fixed by 1050d43c596238630d22d545e6c8de26bdad0ef5

However, I have a few concerns regarding your ssh command. As I understand it, by `ssh my.other.server.net 'cat tomb.key' | tomb open -k - secrets.tomb` you get the tomb key from the server and use it on a tomb locally. I am wrong? This is not what you wrote.

I think a solution using `sshfs` as described [in the wiki](https://github.com/dyne/Tomb/wiki/Tombsshfs) is way better in terms of security.

stevesbrain:
@jaromil - Indeed, I am referring to the new ```-g``` switch :)

@roddhjav - Apologies if my description didn't quite match the command - I just quickly pulled the command from the reference doco. My key lives on my local machine, whilst the tomb lives on the server :) Whilst sshfs would be better in terms of security, I'd still be encountering the issue whereby I have to decrypt locally first (due to the way my keys work). I'll check out the latest commit and see how it goes for me! Thanks

stevesbrain:
So, I've tested it out just locally for the moment, however, unless I'm misunderstanding, it's not working as expected. The command:

```bash
tomb open -k help-gpg.key -g help.me
```

Does work as expected, prompting me for my GPG key password after I've inserted my YubiKey smartcard. However, using gpg to decrypt first, then piping it to tomb, does not work:

```bash
:: [stephen@duck] ~
[...]

Issue #252, opened 2017-03-16
Remove --shared in favor of -r


jaromil:
Is there anything preventing us to remove the --shared flag in favor of a multiple entry detection on the -r flag when in gpg -g mode?
I am wondering since that will simplify the usage of gpg asymmetric keys overall.

roddhjav:
The `--shared` flag is not a technical requirement (tomb does not need this flag to detect the presence of more than one recipient). I implemented this flag only as a ACK from the users they are going to encrypt the tomb key with more than one gpg key. If you think it is not needed, we can get rid of this flag.

We can have a multiple entry detection on the `-r` option instead of separating the key id by a coma in the same -r entry. However, I don't know if tomb support multiple option entry. That said, (in my opinion) it would be a better solution.

jaromil:
I really think is better to remove `--shared` and have comma separated multiple options on `-r`. If something would require conscious confirmation then the `--force` flag is there for that, but I believe this is not a case requiring it.

For parsing comma separated strings there is the `strtok()` function provided by zuper which is available inside Tomb.

jaromil:
thanks, feel free to open a PR when ready with this, or I can merge it by hand from your branch.

roddhjav:
The PR is not ready yet, I'm investigating a few bugs regarding GPG options, I will submit a PR when all this work will be ready.

PR #256, opened 2017-04-05
Options for GnuPG Key - 2

roddhjav:
This PR does three things:
* Remove `--shared`. See #252
*  Support for gpg encrypted tomb key loaded from stdin. See #255
* Fix a few issues regarding GPG default key. 
  - Remove `--no-options` gpg option when using GPG key.
  - Improve gpg default key tests

To use the default key, `~/.gnupg/gpg.conf` needs:
```
default-key <keyid>
default-recipient-self
```
Or
```
default-recipient <keyid>
```
Otherwise the first key in the keyring is used.

They will be other changes in the future regarding wrong use of `-g` option and issue with pinentry. I don't know when they'll be ready, they could be the purpose of an other PR. # #

### DIFF
--- a/extras/test/gnupg/gpg.conf
+++ b/extras/test/gnupg/gpg.conf
@@ -1,0 +1,8 @@
+# ~/.gnupg/gpg.conf
+
+# Description of options can be found here:
+# http://www.gnupg.org/documentation/manuals/gnupg-devel/GPG-Options.html
+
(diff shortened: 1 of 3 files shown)


jaromil:
Great, will test this and run through docs to make sure they are in sync. Take it easy, async key support seems mature enough for a stable release by now.

PR #261, opened 2017-05-12
Missing lsof stuff

Narrat:
PR for the missing stuff since the reintroduction of lsof for slam operation, which means it is now included in --version and a testcase was added.

Thanks to @roddhjav for help with it.

Any missing stuff, or something that should be changed, etc?


jaromil:
Works great, well done Narrat! :+1:

PR #267, opened 2017-06-10
Exec hooks

jaromil:
Small refactoring of `post-hooks` now renamed to `exec-hooks` accepting two callbacks states: `open` and `close` that are passed as string arguments to the script or executable in place. Documentation updated accordingly.


jaromil:
Merging after a period of 3 days for comments. This is a minor breaking change for the `post-hooks` functionality.

PR #275, opened 2017-07-21
Double check after umounting bind directories

gador:
Checking if a /dev/mapper/tomb* is still present in the
output of mount after umounting binded directories.
Addresses issue #265

### DIFF
--- a/tomb
+++ b/tomb
@@ -2662,10 +2662,30 @@ umount_tomb() {
                 }
             }
         done
+        
+        # check if the tomb is actually still mounted. Background:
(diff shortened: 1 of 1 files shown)


gador:
Changed grep's behavior. It now should work, even if multiple tombs are opened.

jaromil:
can you rebase into a single commit?

gador:
Done.

gador:
Do you still need something done for the commit?

jaromil:
thanks!

PR #283, opened 2017-08-17
Use `findmnt` instead of parsing the output of `mount -l`

calvertvl:
The mount-related functionality (finding the volume label) is broken for me on NixOS (unstable) with Tomb 2.4, zsh 5.3.1, and util-linux 2.30.

As of util-linux v2.30 (or earlier), `mount -l` does not list the filesystem label, breaking all of the awk patterns. (Note that on Ubuntu 16.04, which uses util-linux 2.27.1, the parsing still works, so something in a later release of util-linux broke it.)

This means that things like `tomb close <name>` and `tomb list <name>` fail.

The `findmnt` command was introduced in util-linux v2.18 in June 2010.

The manpage for `mount` as of util-linux v2.22 (September 2012) says:
> The listing mode is maintained for backward compatibility only.
> For  more  robust  and  customizable  output use `findmnt(8)`, **especially in your scripts.**

As such, `tomb` should be updated to use `findmnt` instead of `mount`, unless there is a compelling reason otherwise.

This PR vastly simplifies the logic for `list_tomb_binds()`; it works for me on Ubuntu 16.04 and NixOS.

I don't have a Debian VM handy to test on, but hopefully the bugs there don't apply to `findmnt` output&mdash;if one does not already exist, a regression case to confirm proper functionality should be written.

### DIFF
--- a/tomb
+++ b/tomb
@@ -549,7 +549,7 @@ is_valid_tomb() {
     _plot $1     # Set TOMB{PATH,DIR,FILE,NAME}
 
     # Tomb already mounted (or we cannot alter it)
-	[[ "`_sudo findmnt -lvo SOURCE,TARGET,FSTYPE,OPTIONS,LABEL |
+	[[ "`_sudo findmnt -lrvo SOURCE,TARGET,FSTYPE,OPTIONS,LABEL |
         awk -vtomb="[$TOMBNAME]" '
 /^\/dev\/mapper\/tomb/ { if($5==tomb) print $1 }'`" = "" ]] || {
         _failure "Tomb is currently in use: ::1 tomb name::" $TOMBNAME
@@ -2263,7 +2263,7 @@ awk "/mapper/"' { print $2 ";" $3 ";" $4 ";" $5 }'`
 list_tomb_mounts() {
     [[ -z "$1" ]] && {
         # list all open tombs
-        _sudo findmnt -lvo SOURCE,TARGET,FSTYPE,OPTIONS,LABEL \
+        _sudo findmnt -lrvo SOURCE,TARGET,FSTYPE,OPTIONS,LABEL \
             | awk '
 BEGIN { main="" }
 /^\/dev\/mapper\/tomb/ {
@@ -2274,7 +2274,7 @@ BEGIN { main="" }
 '
     } || {
         # list a specific tomb
-        _sudo findmnt -lvo SOURCE,TARGET,FSTYPE,OPTIONS,LABEL \
+        _sudo findmnt -lrvo SOURCE,TARGET,FSTYPE,OPTIONS,LABEL \
             | awk -vtomb="[$1]" '
 BEGIN { main="" }
 /^\/dev\/mapper\/tomb/ {
@@ -2302,7 +2302,7 @@ list_tomb_binds() {
 
     # note that this code assumes that the tomb name is provided in square brackets
 
-    _sudo findmnt -lo SOURCE,TARGET,FSTYPE,OPTIONS,LABEL \
+    _sudo findmnt -lro SOURCE,TARGET,FSTYPE,OPTIONS,LABEL \
             | grep -F "/dev/mapper/tomb" \
             | awk -vpattern="$1" '
 BEGIN { }
@@ -2501,7 +2501,7 @@ resize_tomb() {
     _load_key # Try loading new key from option -k and set TOMBKEYFILE
 
     local oldtombsize=$(( `stat -c %s "$TOMBPATH" 2>/dev/null` / 1048576 ))
-    local mounted_tomb=`_sudo findmnt -lvo SOURCE,TARGET,FSTYPE,OPTIONS,LABEL |
+    local mounted_tomb=`_sudo findmnt -lrvo SOURCE,TARGET,FSTYPE,OPTIONS,LABEL |
         awk -vtomb="[$TOMBNAME]" '/^\/dev\/mapper\/tomb/ { if($5==tomb) print $1 }'`
 
     # Tomb must not be open
--- a/tomb
+++ b/tomb
@@ -2302,7 +2302,8 @@ list_tomb_binds() {
 
     # note that this code assumes that the tomb name is provided in square brackets
 
-    findmnt -lo SOURCE,TARGET,FSTYPE,OPTIONS,LABEL /dev/mapper/tomb* \
+    findmnt -lo SOURCE,TARGET,FSTYPE,OPTIONS,LABEL \
+            | grep -F "/dev/mapper/tomb" \
             | awk -vpattern="$1" '
 BEGIN { }
 {
--- a/tomb
+++ b/tomb
@@ -2302,7 +2302,7 @@ list_tomb_binds() {
 
     # note that this code assumes that the tomb name is provided in square brackets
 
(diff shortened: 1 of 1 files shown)


jaromil:
First of all thanks for the PR! I agree with your analysis and think its necessary.

The tests on travis (running on ubuntu 14) fail on `tomb close`. Do you have an idea why? The `close` command is indeed affected by your changes. If needed I will also test it on a ubuntu 14 trying to reproduce the problem, but that will take a while.

calvertvl:
It pretty much has to be related to the version of util-linux, which for Ubuntu 14.04 is 2.20.1.

It probably is that the `LABEL` column is not parsing as expected; that's the same error message I was getting on my NixOS box where `mount -l` doesn't show the labels.

I'm spinning up VMs for Ubuntu 14.04, 16.04, and 17.04, as well as Debian 7, 8, and 9; I'll try to get answers.

calvertvl:
Found it: `findmnt` does not have `FSROOT` as a column in 2.20.1, so I'm rewriting the AWK to parse it from `findmnt -lo` instead of using `findmnt -lvo` with `FSROOT`.

I will re-test after; hopefully in the next few hours.

calvertvl:
Still seems to not work: `findmnt` 2.20.1, at least on Ubuntu 14.04, does not show labels in the `LABEL` column (or in any other way that is obvious); maybe this was a bug in that version?

Ubuntu 16.04 works fine with these latest changes. I'll test against Debian.

calvertvl:
Ok, here's what we have as results from `make test`:
- [X] NixOS 17.09pre, util-linux 2.30
- [X] Ubuntu 17.04, util-linux 2.29
- [X] Ubuntu 16.04, util-linux 2.27.1
- [ ] Ubuntu 14.04, util-linux 2.20.1, `findmnt` does not show labels
- [ ] Debian 7, util-linux 2.20.1, `findmnt` does not show labels
- [ ] Debian 8, util-linux 2.25.2, `findmnt` does not show labels
- [ ] Debian 9, test failures related to GPG keys (in `20_recipients.sh` at `is_valid_recipients`)

For the ones where `findmnt` does not show labels, it looks like running it as `sudo findmnt` will show the labels, I guess because labels were considered a potential security issue?

The above would be the easy solution, but it would mean that sudo would now be needed for more commands (`tomb list`, `tomb index`, and `tomb search`, possibly more). If that is acceptable, then problem solved...just add `_sudo` before all calls to `findmnt`.

Related: findmnt is used in the following:
* `list_tomb_mounts`, used in `tomb list`, `tomb close`, `tomb index`, and `tomb search`
* `list_tomb_binds`, used in `tomb list` and `tomb close`
* `is_valid_tomb`
* `tomb resize`

Looks like other options that wouldn't change current sudo-related behavior would be:
* When opening a tomb, run `sudo blkid >/dev/null` to update the cachefile, then use `label=$(blkid -s LABEL -o value <device>)` as unprivileged user to pull the labels
    * Note that if there is a problem such that `sudo blkid` does not get run after cryptsetup, the label would be unavailable. This is probably not an issue: just add it after the `tune2fs` call in `mount_tomb()`, and just before the success message in `umount_tomb()`.
    * I'd probably implement this as `_refresh_label_cache() { _sudo blkid >/dev/null }`
* Do some parsing magic on `/dev/disk/by-label/` and `/dev/mapper/tomb.*` (the symlinks would point to the same `/dev/dm-N` device)

Both of these would require reasonably substantial changes to the existing logic, at least in some regards (e.g. awk filters the output based on the label, which would no longer be dependable).

calvertvl:
A couple of things:
* The sudo thing seems to be Ubuntu/Debian, based on comments from others regarding e.g. `lsblk`, though I haven't tested old versions of other distros.
* `findmnt -lr` is preferable to `findmount -l`, as it makes the output more parsing-friendly (no odd linewrapping or truncation in older versions).

By adding `_sudo` to all `findmnt` lines makes Debian 8 pass.

Also adding `-r` to all `findmnt` lines makes Debian 7 and Ubuntu 14.04 pass.

jaromil:
Thanks for investigating the issue extensively! I suspected this would be quite some work.

I'm not sure extending the need for `_sudo` over so many more commands is the right route, yet it feels cleaner. The cache solution is also very interesting. Please give me and others some time to think about this and make some tests.

jaromil:
Hi @calvertvl, I am ready to merge this, but would prefer to do it without the whitespace cleanup for the sake of clarity. Is it easy for you to rebase this PR eliminating the whitespace changes?

jaromil:
I will try to rebase this now, to eliminate the whitespace cleanup while keeping your original commits.

calvertvl:
Hi @jaromil, I was rather busy the last couple of months, and am just
getting back to things now.

If you have trouble, let me know, and I can rework it; it shouldn't take
long for me to rebase if I have to.

jaromil:
ahoy! yea about time i understand very well ;^) I've just rebased it in a single commit now in use-findmnt branch but it errors out, I suspect a small conflict with my recent refactoring of slam(). See https://travis-ci.org/dyne/Tomb/builds/324109767

jaromil:
Also I just noticed the option flags used for findmnt give an error:

```
findmnt from util-linux 2.29.2
$ findmnt -lrvo SOURCE,TARGET,FSTYPE,OPTIONS,LABEL
findmnt: these options are mutually exclusive: --pairs --list --raw --verify
```

calvertvl:
I'm having trouble running the tests on NixOS because zsh is in a non-standard location,
 so the makefile has to use `/usr/bin/env zsh` like everywhere else. I don't remember how I ran the tests on NixOS before...

I'm working on updating it so that tests will run for NixOS on install; hopefully I can get that working.

As for `findmnt`, apparently something has changed, as it worked at the time. The options probably should be `-Arvo` instead of `-lrvo`.

roddhjav:
@calvertvl  This is because of: https://github.com/dyne/Tomb/blob/742b097192c7a9eadd4d3448329f8b9140708fd0/extras/test/Makefile#L20

To run the test with a different zsh  location, in `extra/tests/Makefile` replace:
```
SHELL := /usr/bin/zsh
```
by:
```
SHELL ?= $(which zsh)
```

calvertvl:
@roddhjav, that works, but is there a way to use `/usr/bin/env zsh` like everywhere else, for consistency?

roddhjav:
@calvertvl The header for all the test files is already `/usr/bin/env zsh`. However, it does not matter because all the files are either executed by the makefile using this `$SHELL` variable or sourced.

calvertvl:
@jaromil, looks like you fixed it?

Your `use-findmnt` branch seems to match with my test rebase, and in both cases, all tests pass for me on NixOS except the regression tests; those failures are expected (because of the label detection not working on NixOS).

And @roddhjav, thanks for the comments; the big issue for the NixOS automated tests is that builds are sandboxed and don't have access to `/tmp` so e.g. `/tmp/zsh` can't be  created. I'm pretty sure that's a NixOS issue (and in any case is unrelated to this PR), so I'll work on that separately.

jaromil:
Ok your PR is succesfully rebased and merged with the conflicts. This was it:
https://github.com/dyne/Tomb/commit/b72d67618bbae537cb2eaaba2a6ece1a3e475bdd
a fix I've introduced in the last slam() refactoring.
I'll open a new PR and we can close this.

jaromil:
many thanks Victor and welcome among Tomb's contributors!

PR #284, opened 2017-10-07
Fix recover_key and add key recovery during exhume

privb0x23:
- Fixed recover_key to remove the KDF header (first line) before insertion between header/footer
- Added key recovery (header/footer placed around the encrypted key) when extracting from an image file

Signed-off-by: privb0x23 <privb0x23@users.noreply.github.com>

### DIFF
--- a/tomb
+++ b/tomb
@@ -923,7 +923,11 @@ recover_key() {
 
     TOMBKEY=""        # Reset global variable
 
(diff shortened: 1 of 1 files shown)


jaromil:
hi there, thanks for contributing!

Unfortunately this PR breaks usage of steganographic keys when pulled from stdin (see https://travis-ci.org/dyne/Tomb/builds/284823475)

Also I lack good understanding of the problem you are solving, can you explain it?

PR #290, opened 2017-12-06
use shred instead of wipe

parazyd:
This commit will use `shred` (which is part of util-linux) instead of `wipe`, which is yet another external dependency.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,7 @@ all:
 	@echo "To install it in /usr/local together with its manpage use 'make install'."
 	@echo
 	@echo "To run Tomb one needs to have some tools installed on the system:"
-	@echo "Sudo, cryptsetup, pinentry and gnupg. Also wipe is recommended."
+	@echo "Sudo, cryptsetup, pinentry and gnupg."
 	@echo
 
 install:
--- a/tomb
+++ b/tomb
@@ -781,7 +781,7 @@ function _print() {
 
 _list_optional_tools() {
     typeset -a _deps
-    _deps=(gettext dcfldd wipe steghide)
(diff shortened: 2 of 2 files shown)


Narrat:
I would also favour this change, but as seen in the past, changes must run across various distributions.
So it should be made sure this change works with all current supported distributions (like Debian, Fedora)
And sure about util-linux? On Arch it comes with coreutils

Narrat:
Note regarding this PR.
Shouldn't the shred command be extend the -u option? So it will remove the mangled file, as shred doesn't do that by default (but I assume wipe is doing?)

jaromil:
Thanks for the review Narrat. My thoughts:
1. The **Please Note** part in `shred`'s manual applies also to `wipe` IMO
2. Do we have an overview of the parts in which `$WIPE` is used in Tomb?

parazyd:
@Narrat My bad. It's indeed from coreutils. Also rebased for proper shred params. About reworking `$WIPE`, I don't think it's an option yet since some distributions do not come with GNU Coreutils (e.g. Alpine).

@jaromil `$WIPE` is used once in `_endgame()` to clear temporary files. Nowhere else.

Narrat:
Ah, right. Different implementations.
BSD would also be a candidate without shred, until coreutils is installed (gshred).
What is Alpine using? I see the gnu coreutils package is available, but fairly outdated and not including shred.
But funny to see a Rust based coreutils, which does contain shred :D

jaromil:
Ahoy folks, are we good with this? I am.

Narrat:
Totally good with it.

PR #296, opened 2018-01-02
Replace 'mount -l' with 'findmnt'

jaromil:
This fixes a mount-related functionality (finding the volume label) in new versions of util-linux, that since v2.30 does not list anymore volume labels with its mount -l command. Since findmnt needs sudo to list labels, this also introduces the need for sudo in more commands: is_valid_tomb(), list, index and search. The issue was examined in PR #283 and this is a rebase of it.

### DIFF
--- a/tomb
+++ b/tomb
@@ -2274,8 +2274,12 @@ BEGIN { main="" }
 '
     } || {
         # list a specific tomb
+		# add square parens if not present (detection)
+		local tname
+		if [[ "${1[1]}" = "[" ]]; then tname="$1"
+		else tname="[$1]"; fi
         _sudo findmnt -rvo SOURCE,TARGET,FSTYPE,OPTIONS,LABEL \
-            | awk -vtomb="[$1]" '
+            | awk -vtomb="$tname" '
 BEGIN { main="" }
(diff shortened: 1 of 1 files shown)


jaromil:
Waiting for tests and last minute reviews, then will rebase my fixes in one single commit and merge this.

Narrat:
What I don't like is the increased usage of sudo. Especially since the first approach in the other PR would have worked for me :D Too bad findmnt works totally different between versions (or distribution specific patches are to blame, dunno).
But yeah, using findmnt is cleaner.
Didn't take a look yet at your additional changes

Narrat:
Any issues without the square parens detection, or what is the rationale for going back?

calvertvl:
@Narrat, some platforms use square parens and others do not. The awk script to retrieve the label assumes that they are present.

jaromil:
Now erroring for the 'A' flag. Wow, findmnt changes **a lot** between versions. I'm wondering if this really is more reliable than reading mtab. Sure it's meant to, but really?!
```
findmnt: invalid option -- 'A'
```
I will remove this last option and include its version in `tomb --version`

PR #298, opened 2018-01-03
Fixes: version check & tests

roddhjav:
Hi @jaromil 

As it seems you are going to do a release soon, there are a few fixes I have on my local system and that can be merged quite easily:
* Use the zsh function `is-at-least` instead of `=~` to check program version.
* Support non-standard zsh location in the test suite. See #283
* Remove useless cleanup in test setup.

Regards,
Alex

### DIFF
--- a/extras/test/40_regression.sh
+++ b/extras/test/40_regression.sh
@@ -3,10 +3,11 @@
 export test_description="Tomb regression tests"
 
 source ./setup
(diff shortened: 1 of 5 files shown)


jaromil:
Awesome! and yes, good intuition. Most standing issues are solved so its time for a 2.5
Nice the `is-at-least` function! did not know it.

Issue #300, opened 2018-01-05
Complete output of 'pinentry --version' needed?


Narrat:
pinentry --version (1.1.0) prints the following:
>pinentry-gtk2 (pinentry) 1.1.0
Copyright (C) 2016 g10 Code GmbH
License GPLv2+: GNU GPL version 2 or later <https://www.gnu.org/licenses/>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.

Is it necessary to display the License info? If not, maybe |head -n1 could be added so only the version is printed in tomb -v

Edit:
else it looks like that:
```
  System utils:
  
  Sudo version 1.8.21p2
  cryptsetup 2.0.0
  pinentry-gtk2 (pinentry) 1.1.0
Copyright (C) 2016 g10 Code GmbH
License GPLv2+: GNU GPL version 2 or later <https://www.gnu.org/licenses/>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
  findmnt from util-linux 2.31.1
  gpg (GnuPG) 2.2.4 - key forging algorithms (GnuPG symmetric ciphers):
  /usr/bin/gpg
```

Issue #299, opened 2018-01-05
list_gnupg_ciphers() fails with non english locales


Narrat:
The output of gpg --version is localized, so the search for Cipher may fail and provide no output in tomb -v.
Dunno atm how to handle this, as there doesn't seem to be a --list-ciphers like option in gpg?
And dunno how nice it is, to prepend the gpg command with LANG=C

jaromil:
does it works by prepending LANG=C inside the backtick at line 1365?

Narrat:
Yeah, it does work, but is it a nice way to solve this in such a way? (Else I would have created a PR.)

But imo the whole thing could see some love?
Like there is a small formatting issue in the output (missing a space in front), and the construct itself. Does it really make sense to include the binary path? And if it make sense it should be placed differently. It isn't a key forging algo.

```
  gpg (GnuPG) 2.2.4 - key forging algorithms (GnuPG symmetric ciphers):
  /usr/bin/gpg
 IDEA 3DES CAST5 BLOWFISH AES AES192 AES256 TWOFISH CAMELLIA128 CAMELLIA192 CAMELLIA256
```
Maybe all (hard deps and opt deps) should display the version (if found), display only the binary path or both. Personally I would favor displaying the version.

jaromil:
I'd prefer less cruft but you are welcome to make a PR then!

PR #301, opened 2018-01-07
Small things

Narrat:
On my trial to create a zsh_completion file (a better one than I maintain in the AUR tomb-git package at least) I noticed some stuff (see #299 and #300 and some more which I need to address yet).

So some small commits for various things and up to debate.
Commit 2 and 4 could be combined as they touch the same subject, which I will do, if commit 4 is accepted.
Or 2,3 and 4 combined in one commit as a small overhaul of list_gnupg_ciphers(). Whatever is prefered.

Edit: Ah, and against suggesting changes on _list_optional_tools() I won't touch it anytime soon. Some commands only know `-v`, some only the long-option `--version` and some don't know about either of those two.
And changing it to a copy of hard-deps version checking isn't worth time spending on. Although it would be nice to see those version numbers.

### DIFF
--- a/tomb
+++ b/tomb
@@ -1360,7 +1360,7 @@ gen_key() {
 # prints an array of ciphers available in gnupg (to encrypt keys)
 list_gnupg_ciphers() {
 	# prints an error if GnuPG is not found
-	which gpg 2>/dev/null || _failure "gpg (GnuPG) is not found, Tomb cannot function without it."
+	which gpg &>/dev/null || _failure "gpg (GnuPG) is not found, Tomb cannot function without it."
 
 	ciphers=(`gpg --version | awk '
 BEGIN { ciphers=0 }
@@ -1368,7 +1368,7 @@ BEGIN { ciphers=0 }
 /^Hash:/ { ciphers=0 }
 { if(ciphers==0) { next } else { gsub(/,/,""); print; } }
 '`)
-	print " ${ciphers}"
+	print "${ciphers}"
 	return 1
 }
 
--- a/tomb
+++ b/tomb
@@ -1362,10 +1362,12 @@ list_gnupg_ciphers() {
 	# prints an error if GnuPG is not found
 	which gpg &>/dev/null || _failure "gpg (GnuPG) is not found, Tomb cannot function without it."
 
+	# On gpg1 and gpg2 line 10 always point to available ciphers.
+	# Print those until the next section is found
 	ciphers=(`gpg --version | awk '
 BEGIN { ciphers=0 }
-/^Cipher:/ { gsub(/,/,""); sub(/^Cipher:/,""); print; ciphers=1; next }
-/^Hash:/ { ciphers=0 }
+NR==11 { gsub(/,/,""); sub(/^.*:/,""); print; ciphers=1; next }
+/^.*:/ { ciphers=0 }
 { if(ciphers==0) { next } else { gsub(/,/,""); print; } }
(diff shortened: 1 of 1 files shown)


tomb
```diff
@@ -1362,10 +1362,12 @@ list_gnupg_ciphers() {
 	# prints an error if GnuPG is not found
 	which gpg &>/dev/null || _failure "gpg (GnuPG) is not found, Tomb cannot function without it."
 
+	# On gpg1 and gpg2 line 10 always point to available ciphers.
+	# Print those until the next section is found
 	ciphers=(`gpg --version | awk '
 BEGIN { ciphers=0 }
-/^Cipher:/ { gsub(/,/,""); sub(/^Cipher:/,""); print; ciphers=1; next }
-/^Hash:/ { ciphers=0 }
+NR==11 { gsub(/,/,""); sub(/^.*:/,""); print; ciphers=1; next }
```
jaromil: How can we be sure that the line listing the ciphers is always 11 in previous and future versions of GnuPG?
I think a safer bet is to assume the line is prefixed with Cipher:

jaromil:
Thanks! I find all very acceptable but the third (5e3b0de) where I've made a comment.
Unfortunately I have no better solution to offer, so if you see no alternative I am fine with it.

tomb
Narrat: Yeah, not easy to decide what is the best.
If we go with Cipher, then we need to force locale with LANG=C, which could fail if LC_ALL or LC_COLLATE is set. And I kinda dislike such a solution, as it sets on a fixed setting (although in this case it might be forgiveable).
Then there could be some sed magic, but dunno if this will be easier.

Regarding previous and future versions:
I looked into the git history.
Two functions are important.
1.  show_version() from util/argparse.c . Comparing this function against the latest gpg1 release it was like that since the [GPL3 switch](https://git.gnupg.org/cgi-bin/gitweb.cgi?p=gnupg.git;a=blobdiff;f=util/argparse.c;h=601f6660ef3aaed01e11538514f4a2334981fe17;hp=cd29c0cef6de4e61a2b13f957cad6bb721b25abb;hb=9a2a818887b4d36ca5aed23fb0381e0e02653b6a;hpb=960ac800487dbc9d91bbef005fee4ecbfae72e83) done 2007 (so earliest gpg1 that would work: good question. Earliest visible tag is 1.4.14 from 2013. But I assume there were some versions before). What was missing in the previous implementations was the License info.

1. strusage() from g10/gpg.c. Should be working with earlier implementations as well, but the latest gpg1 matches the one from [2005](https://git.gnupg.org/cgi-bin/gitweb.cgi?p=gnupg.git;a=blob;f=g10/gpg.c;h=10eb24c413a2a4a43b31f1e01867e7e606c618fe;hb=bd1df0119c26d14e298d11abdd07ae77e8fb124f).

Tested the fixed linenumber with gpg1 and gpg 2.2.4. So, fingers crossed, I don't assume there will be changes anytime soon.

That would my arguments :D Your call :P

Narrat:
In addition to my comment on the commit
What else could work? Dunno. Every other implementations has something ugly to offer which could make it fail in one way or other

jaromil:
yes, ok. let's please try to keep an eye if the 11 line changes through versions.

Narrat:
Yeah, will have a look at this

jaromil:
cheers. these are small details but alll counts. I appreciate your attention to them

PR #303, opened 2018-01-28
Added pinentry-qt5 support

sargo-devel:
Please add support for pinentry-qt5. The pinentry-qt4 is not used anymore in some distros.

### DIFF
--- a/tomb
+++ b/tomb
@@ -377,7 +377,7 @@ ask_password() {
 	# Distributions have broken wrappers for pinentry: they do
 	# implement fallback, but they disrupt the output somehow.	We are
 	# better off relying on less intermediaries, so we implement our
-	# own fallback mechanisms. Pinentry supported: curses, gtk-2, qt4
+	# own fallback mechanisms. Pinentry supported: curses, gtk-2, qt4, qt5
 	# and x11.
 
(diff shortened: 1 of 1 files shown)


tomb
```diff
@@ -377,7 +377,7 @@ ask_password() {
 	# Distributions have broken wrappers for pinentry: they do
 	# implement fallback, but they disrupt the output somehow.	We are
 	# better off relying on less intermediaries, so we implement our
-	# own fallback mechanisms. Pinentry supported: curses, gtk-2, qt4
+	# own fallback mechanisms. Pinentry supported: curses, gtk-2, ,qt4, qt5
```
Narrat: Just a cosmetic nitpick

Narrat:
General question @jaromil and co. Isn't it possible to reduce the code duplication? Every output that is produced contains the same vars so I'm wondering if there is a reason it was done this way

tomb
sargo-devel: Ok. Done.

jaromil:
indeed, we could just avoid duplication of the assuan input sequence to pinentry into a single function. Will do that after merging this, thanks all for the suggestions!

PR #305, opened 2018-01-29
put assuan commands to pinentry into a single function

jaromil:
avoid code duplication since all pinentry impelmentation need the same sequence of assuan commands, according to suggestion in #303 by @Narrat


Narrat:
Looking good and tomb is still working :D 👍

Issue #307, opened 2018-02-03
Security issue when using tomb + kdf


AitorATuin:
When using KDF support, passwords containing whitespaces are trimmed out without even notifying the user (in any case since without KDF users can have passwords containing whitespaces, when using KDF users should be able to use them as well). 

Easy to reproduce, see this snippet:

```
% salt=$(./bin/tomb-kdb-pbkdf2-gensalt)
% iter=$(./bin/tomb-kdb-pbkdf2-getiter)

% ./bin/tomb-kdb-pbkdf2 ${salt} ${iter} 64 2>/dev/null <<<'one two three'
fd144a3923ed7bc5a5e04a1b7ec4d69d1edb71da5aa9f0acc4715beeb51690c8b5c300cc47bcb6b05ccde1447d24a5a408c13362bdee167702f939763ab93120

% ./bin/tomb-kdb-pbkdf2 ${salt} ${iter} 64 2>/dev/null <<<'one crappy shit here'
fd144a3923ed7bc5a5e04a1b7ec4d69d1edb71da5aa9f0acc4715beeb51690c8b5c300cc47bcb6b05ccde1447d24a5a408c13362bdee167702f939763ab93120

% ./bin/tomb-kdb-pbkdf2 ${salt} ${iter} 64 2>/dev/null <<<'one'
fd144a3923ed7bc5a5e04a1b7ec4d69d1edb71da5aa9f0acc4715beeb51690c8b5c300cc47bcb6b05ccde1447d24a5a408c13362bdee167702f939763ab93120
```

As you can see passwords are 'parsed' before inputting them into `gcry_kdf_derive` function.

The problem seems to be exactly here: https://github.com/dyne/Tomb/blob/master/extras/kdf-keys/pbkdf2.c#L108

`fscanf` should be avoided!

Possible solutions:

Using fgets instead:

`fgets(pass, BUF_SIZE, stdin);`

Reading char by char from the input:
```
   int j = 0;
   while (j < (BUF_SIZ + 1)) {
       char c = getchar();
       if (c == EOF) break;
       pass[j] = c;
       j++;
   }
   if (j == BUF_SIZ + 1) {
       fprintf(stderr, "Error: password is too long\n");
       exit(1);
   }
   pass[j-1] = '\0';
```

boyska:
I wrote that file in 2e6a3d, and I can confirm that the bug you reported is true, as I should have noted reading `man scanf`, which, describing the `%s` modifier says:

[...]

PR #308, opened 2018-02-03
Fixes a problem reading passwords when using the kdf wrapper  

AitorATuin:
Fixes this issue: https://github.com/dyne/Tomb/issues/307

There was also a test in `extras/kdf-keys` that was failing (it was assuming that the NULL byte could be part of the password), this PR makes the test to pass as well.

Some tests have been added checking that passwords are not split if they have whitespaces and checking len limits for passwords

It also enables the execution of tests in `extras/kd-keys` in travis' build

### DIFF
--- a/extras/kdf-keys/pbkdf2.c
+++ b/extras/kdf-keys/pbkdf2.c
@@ -43,8 +43,8 @@
 
 #include <gcrypt.h>
 
-/* Max password size */
-#define BUFFER_SIZE 1024
+/* block size for password buffer */
+#define BLOCK_SIZE 40
 
 /* TODO: move print_hex and hex_to_binary to utils.h, with separate compiling */
 void print_hex(unsigned char *buf, int len)
@@ -76,15 +76,37 @@ int hex_to_binary(unsigned char *buf, char *hex)
 	return count;
 }
 
+void cleanup(char *result, int result_len, char *pass, char *salt, int salt_len) {
+	int i;
+
+	//clear and free everything
+	if (result) {
+		for(i=0; i<result_len;i++)
+			result[i]=0;
+		free(result);
+	}
+	if (pass) {
+		for(i=0; i<strlen(pass); i++) //blank
+			pass[i]=0;
+		free(pass);
+	}
+	if (salt) {
+		for(i=0; i<salt_len; i++) //blank
+			salt[i]=0;
+		free(salt);
+	}
+}
+
 int main(int argc, char *argv[])
 {
-	char pass[BUFFER_SIZE];
-	unsigned char *salt;
+	char *pass = NULL;
+	unsigned char *salt = NULL;
 	int salt_len;                  // salt length in bytes
 	int ic=0;                        // iterative count
 	int result_len;
-	unsigned char *result;       // result (binary - 32+16 chars)
-	int i, pw_lenght;
+	unsigned char *result = NULL;       // result (binary - 32+16 chars)
+	int i, pw_len = 0;
+	int buff_len = BLOCK_SIZE;
 
 	if ( argc != 4 ) {
 		fprintf(stderr, "usage: %s salt count len <passwd >binary_key_iv\n", argv[0]);
@@ -92,7 +114,8 @@ int main(int argc, char *argv[])
 	}
 
 	//TODO: move to base64decode
-	salt=calloc(strlen(argv[1])/2+3, sizeof(char));
+	salt_len = strlen(argv[1])/2+3;
+	salt = calloc(salt_len, sizeof(char));
 	salt_len=hex_to_binary(salt, argv[1]);
 	if( salt_len <= 0 ) {
 		fprintf(stderr, "Error: %s is not a valid salt (it must be a hexadecimal string)\n", argv[1]);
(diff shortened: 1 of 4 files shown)


extras/kdf-keys/pbkdf2.c
```diff
@@ -43,6 +43,9 @@
 
 #include <gcrypt.h>
 
+/* Max password size */
+#define BUFFER_SIZE 1024
```
Narrat: Why hardcoding the length of the array? (Disregarding the argument, that nobody will have a PW with 1023 chars :D).
Just curious, as this seems a case, where dynamic allocation would be used

extras/kdf-keys/pbkdf2.c
```diff
@@ -105,9 +108,18 @@ int main(int argc, char *argv[])
 		exit(1);
 	}
 
-	fscanf(stdin, "%ms", &pass);
-	if ( pass[strlen(pass)-1] == '\n' )
-		pass[strlen(pass)-1] = '\0';
+	int j = 0;
```
Narrat: Wouldn't it made more sense to name this var something like pw_lenght, or else?
Passing `j-1` to gcry_kdf_derive isn't descriptive, without further looking?

extras/kdf-keys/pbkdf2.c
```diff
@@ -105,9 +108,18 @@ int main(int argc, char *argv[])
...
+	while (j < (BUFFER_SIZE + 1)) {
+		char c = getchar();
+		if (c == EOF) break;
+		pass[j] = c;
+		j++;
+	}
+	if (j == BUFFER_SIZE + 1) {
```
Narrat: Wouldn't this fail if j is +2 than Buffer_Size?

extras/kdf-keys/pbkdf2.c
```diff
@@ -111,19 +111,24 @@ int main(int argc, char *argv[])
...
 	 */
-	int j = 0;
 	while (j < (BUFFER_SIZE + 1)) {
 		char c = getchar();
 		if (c == EOF) break;
 		pass[j] = c;
 		j++;
 	}
-	if (j == BUFFER_SIZE + 1) {
+	if (j >= BUFFER_SIZE + 1) {
```
Narrat: Shouldn't the check be on BUFFER_SIZE? 
Should the password have 1024 chars, it will loose one char, when it is made sure that the string array ends with `\0`

Narrat:
I learned recently, that there is at least for scanf the **special Edit set Conversion Code**. Is this available for fscanf too, and could it be an alternative regarding possible whitespaces in passwords? (Again just curiosity, as I would propably avoided the use of fscanf related functions. Maybe you know more about this)

extras/kdf-keys/pbkdf2.c
AitorATuin: 👍 It makes sense for sure

AitorATuin:
> I learned recently, that there is at least for scanf the special Edit set Conversion Code. Is this available for fscanf too, and could it be an alternative regarding possible whitespaces in passwords? (Again just curiosity, as I would propably avoided the use of fscanf related functions. Maybe you know more about this)

Sure! it supports it. Actually my first check to make sure that the problem was originated by `fscanf` was to use a set with `fscanf` letting it to get spaces as well; that confirmed the source of the bug.  

The problem i see is that `scanf` functions are rather magical and can alter user's data without the user knowing it (exactly what was happening here). For example, using the set allowing spaces was not fixing the problem with the NULL byte (there was a test in `test.txt` failing because of that).

I think the wrapper should not try to be clever than the user when reading the password, so getting in whatever the user is giving to it looks like a better approach (I could be wrong though).

extras/kdf-keys/pbkdf2.c
AitorATuin: Actually the max password is BUFFER_SIZE - 1. 
Anyway I will double check it.

extras/kdf-keys/pbkdf2.c
AitorATuin: Sure that no one will have a password with 1024 characters :) That's the reason i set it up to that size. Just that I thought that it's simpler to have a predefined size and to check for known limits. 
We could though use dynamic allocation if we get the size of the password as a parameter, what do you think about that?

extras/kdf-keys/pbkdf2.c
AitorATuin: I think the logic works, i added a test checking that a password of 1023 characters is accepted and that 1024 characters is refused.

Anyway i removed the `+ 1` fixing the equality operators.

extras/kdf-keys/pbkdf2.c
Narrat: Yeah, I suppose it could be reduced, although who knows :D Maybe there are people with really long passwords.
Regarding dynamic allocation I can think of three setups, but dunno about costs or good practice
  * Set BUFFER to 20/30/40 or sth like that and allocate array. Run through the reading process, and if BUFFER is exceeded, realloc with 2*BUFFER and so on.
  * No BUFFER at all. Reading password and char for char a realloc (although I suppose this will be the slowest and the logic behind.. well)
 * Process password, deduce length, set array and process again (similar stupid like the second, because two times the work)

extras/kdf-keys/pbkdf2.c
Narrat: Ah, right. Didn't take the while loop into account

Narrat:
At least it sounds reasonable to me ;) :D

jaromil:
hi @AitorATuin and many thanks for spotting the bug and packing this PR! Next step is assessing the impact of the fix on tombs that are already in use, which I think is limited to KDF passwords with whitespace. Nevertheless the documentation should reflect this in KNOWN_BUGS. One note on the last variable renaming, which is a good idea indeed: the correct english spelling is `length` not `lenght` (the `h` is in last position).

BTW, slightly off-topic, we may want to evolve this code to support also Argon KDF which is much more reliable against brute-forcing.

AitorATuin:
Thanks to you for tomb, I like the idea.

The problem seems to be only in kdf using passwords with spaces so anyone
having tombs with that password scheme should change them until the fix is
merged and in production.

Tombs using gpg keys or using just passwords to encrypt the key should not
be affected since they don't use the kdf wrapper at all.

Nice catch the typo, I think that in the last commit I have changed all the
length suffixes just to len to keep it coherent with other variable names.
Anyway I will check it tomorrow.

Narrat:
Any reason this is still unmerged?
Although before merge it would propably best if this thing is squashed. Maybe the enhancement of the tests (https://github.com/dyne/Tomb/pull/308/commits/bce58cae3e9e2473547c1f6bb4bc33cad93202e4) could be separate + integration with travis.
But everything else should be in one commit.

Noticed, that the ` extras/kdf-keys/test.sh` changed from 100644 to 100755. Should also be changed back.

jaromil:
I'll merge this, its fine. I'm just looking for some time to write the consequences of this fix on kdf passwords with spaces inside the KNOWN_BUGS file

Narrat:
It would fit the permissions of the other test files, as the stuff is run with runtests. But as jaromil seems to be fine with it, no need to change anything :)

jaromil:
Thanks everyone and apologies for the delay, I'm finally merging this and considering a new 2.6 release.

PR #309, opened 2018-02-17
Fix mounted volume default path

Ganondolf:
Default behavior for tomb is to mount the volume filename.tomb in /media/filename.

Ubuntu 16.04, Tomb 2.5


jaromil:
thanks! I've left the gtk-tray a bit behind since I'm not using it anymore and since Ubuntu dropped support for the normal tray API in 14 series... is that back? what a mess they have made with compatibilities...

Issue #313, opened 2018-05-25
Percent sequences seem to be interpreted when using --tomb-pwd


Pawamoy:
I think this issue is related to #304, because it's about handling passwords containing backslahes or percent signs.

This issue however only happens when using the `--tomb-pwd` (and therefore the `--unsafe`) option:
my password contains `%A`, and when trying to open the tomb from a script with `--unsafe` and `--tomb-pwd` options, the `%A` simply disappears. I suspect the culprit to be `printf -v` here :slightly_smiling_face: 

```
tomb [D] tomb-pwd = <REDACTED>
tomb  .  A password is required to use key <REDACTED>
tomb [D] ask_key_password with tombpass: <REDACTED>
tomb [D] get_lukskey
tomb [D] Created tempfile: /tmp/zsh/8713206851490129586
tomb [D] gpg: AES256 encrypted data
tomb [D] [GNUPG:] NEED_PASSPHRASE_SYM 9 3 2
tomb [D] gpg: encrypted with 1 passphrase
tomb [D] [GNUPG:] BEGIN_DECRYPTION
tomb [D] [GNUPG:] DECRYPTION_INFO 2 9
tomb [D] [GNUPG:] DECRYPTION_FAILED
tomb [D] gpg: decryption failed: bad key
tomb [D] [GNUPG:] END_DECRYPTION
tomb [D] get_lukskey returns 1
tomb  .  A password is required to use key <REDACTED>
tomb [D] asking password with tty=/dev/pts/1 lc-ctype=en_US.UTF-8
tomb [D] using pinentry-gtk2
```

jaromil:
Hi Pawamoy, thanks for this result. Is indeed useful to pin down the problem.
Zsh manual indicates the existance of two interesting flags:
```
-r
Ignore the escape conventions of echo.

-R
Emulate the BSD echo command, which does not process escape sequences unless the -e flag is given. The -n flag suppresses the trailing newline. Only the -e and -n flags are recognized after -R; all other arguments and options are printed.
```

I wonder if `-r` or most likely `-R` would help here if used preceeding `-n` in the `option_value` implementation at line 688. Can you test it? seems you found a pretty easy way to replicate.

Pawamoy:
In fact `print` is innocent. The next suspect is `zparseopts`: if I add a `_verbose "${OPTS[--tomb-pwd]}"` right after the options parsing, I see that the `%A` part of the password is already missing.

Pawamoy:
[...]

PR #314, opened 2018-05-28
Fix issue 313: unescape pinentry %25 to %

Pawamoy:
A simple fix for #313. Using `password="${password//%25/%}"` does not seem to work (zsh?).  There might be a better way than using `sed`. Opening this PR to get feedback :slightly_smiling_face: 

More context: pinentry seems to "unescape" incorrectly the `%` signs, which end up being `%25` in pinentry's output and therefore in tomb's `ask_password()`'s `password` variable.

If the user always use pinentry, then everything works transparently (wrongly but transparently). Now if the same user tries to open the tomb with the `--tomb-pwd` password, it will not work anymore (no pinentry, no `%25`).

I still wonder if it happens to anyone else than me :grin: Does nobody use `%` in passwords :grinning:?

Take care if/when merging such a fix (if a fix is actually needed), as it would prevent users to open tombs with passwords containing `%`.


jaromil:
Right. Still on my mind, been so busy on other shores in the past months...

PR #317, opened 2018-08-18
Add support for multiple GPG subkeys 

bjonnh:
When running (with patch applied):
tomb dig your.tomb -s 10
tomb forge your.tomb -k your.tomb.key -g -r **KEYID1!**,**KEYID2!**
(the ! is to tell gpg to use that specific subkey, not the one it likes)

We now get: (line 1245 in the patch)
<SNIP>
tomb [W] The fingerprints are:
tomb [W] **FINGERPRINT MASTERKEY ** (**KEYID1!**)
tomb [W] **FINGERPRINT MASTERKEY ** (**KEYID2!**)

And gpgopt contains (in the call on the patched line 1325): 
--encrypt --recipient **KEYID1!** --recipient **KEYID2!**

whereas without the patch ${gptopt[@]} is:
--encrypt

So it only uses the default key…

Note: The second commit is to correct a typo I made on line 1245 that included ($gpg_id) inside the call for _fingerprint, which obviously did nothing.

### DIFF
--- a/tomb
+++ b/tomb
@@ -1242,7 +1242,7 @@ gen_key() {
 			_warning "It is your responsibility to check these fingerprints."
 			_warning "The fingerprints are:"
 			for gpg_id in ${recipients[@]}; do
-				_warning "	  `_fingerprint "$gpg_id"`"
+				_warning "	  `_fingerprint "$gpg_id" ($gpg_id)`"
 			done
 
 			gpgopt+=(`_recipients_arg "$recipients_opt" $recipients`)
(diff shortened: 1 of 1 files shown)


bjonnh:
I hope this now clarifies the situation.

jaromil:
ACK. It looks good for me! happy to see subkeys supported. @roddhjav is also fine we can merge this, but it still misses test coverage to avoid regressions and documentation. I can take care of the latter.

roddhjav:
I tested it deeply, I can confirm it works fine (it might even fix possible unknown bug). Therefore @jaromil can merge it.

My only (minor) concert is regarding the output tomb provides when it does not handle subkey, it's redundant, see for example:
```
tomb [W] You are going to encrypt a tomb key with 2 recipient(s).
tomb [W] It is your responsibility to check these fingerprints.
tomb [W] The fingerprints are:
tomb [W] 	  A485 7CD1 76B3 1435 F970 9D25 F0E5 73B8 2894 39CD  (A4857CD176B31435F9709D25F0E573B8289439CD)
tomb [W] 	  0B22 35E6 6075 3AB0 475F B3E2 3DC8 3648 1F44 B31E  (0B2235E660753AB0475FB3E23DC836481F44B31E)
```

Regarding the tests coverage, they are ready, I will propose a PR to add them once this PR is merged.

Issue #266, opened 2017-06-08
When using '-k image.jpeg' the image's 'Date Accessed' attribute should NOT be modified [feature request]


tombraidr:
It's not difficult to find keys hidden in JPEG files -even if you have zillions of them in your system- . You only need search for and sort JPEG images by 'date accessed' in descending order to see a list of potential keys.  

It would be great if when using `tomb open -k path/to/image.jpeg secret.tomb` something behind the scenes triggers something like:
`datetime = getDateAccessed(path_to_image, image.jpeg)` (pseudo code)
and then...
`touch -a -m -t [datetime] image.jpeg`

That way it will give the impression the image hasn't been accessed recently.

What do you think?

arthaud:
That would be great. +1

jaromil:
Yes, I think you spotted another practical vulnerability in the steganographic mechanism of Tomb, well done! we shall address this either within Tomb or within steghide. I am not sure how to best implement the pseudo-code to retrieve the date in shell, perhaps using stat(1)

roddhjav:
I did a implementation of this proposal using `stat` and `touch`. In order to be sure access time is restored in all uses cases and in order to keep the code clean, the time is restored in the endgame function.

See: https://github.com/roddhjav/Tomb/commit/1c5852f0a1d6300b152258f88005d1aee6394526
Please note it has not been tested yet.

tombraidr:
Thanks @jaromil , @roddhjav 

As soon as people say the commit is working for them I shall close the issue.

tombraidr:
@roddhjav 

Ops! I forgot to clarify the `touch -a -m -t` command I wrote above (probably you know already but it's nice to remind us) `-a` is for 'date accessed' and `-m` is for 'date modified'. Both should be considered separately when restoring datetimes.

jaromil:
@roddhjav would you like to pack your commit into a PR?

PR #320, opened 2018-09-23
Track access/modification time of sensitive files

roddhjav:
Collects the stats of tomb keys and tomb files then restore them when
Tomb exits. Can be extended to any file opened by Tomb. See #266

### DIFF
--- a/tomb
+++ b/tomb
@@ -107,6 +107,7 @@ typeset -H TOMBTMP			  # Filename of secure temp just created (see _tmp_create()
 
 typeset -aH TOMBTMPFILES	  # Keep track of temporary files
 typeset -aH TOMBLOOPDEVS	  # Keep track of used loop devices
+typeset -A TOMBFILESSTAT      # Keep track of access date attributes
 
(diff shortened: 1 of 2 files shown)


tomb
```diff
@@ -185,6 +189,27 @@ _is_found() {
...
+_track_stat() {
+    local file="$1"
+    local stat=$(stat --format="%X:%Y" "$file")
+    TOMBFILESSTAT+=("$file" "$stat")
+}
+
+# Restore files stats
+_restore_stat() {
+    local file stat
+    for file stat in "${(@kv)TOMBFILESSTAT}"; do
```
jaromil: nice! I did not know destructuring key/values in iterations over maps worked this way also in zsh.
I have restarted the build on the CI BTW, investigating why the error.

roddhjav:
Just fixed a bug that should not happen.

I actually don't know why the build crashed.

jaromil:
well done! never mind the build, travis is not so-so deterministic unfortunately

PR #331, opened 2018-11-01
Overhaul message printing

jcrd:
Change to print messages to stdout by default, and stderr if the return code for `_msg` is non-zero.
Also check if the stream being printed to is a tty and otherwise disable escape sequences.
Update: the commit effecting the following was excluded from this PR as it caused tests to fail.
> Finally, change `_print "\000"` to `_print ""` as `less` (for example) doesn't like the former and they seem to be equivalent.

### DIFF
--- a/tomb
+++ b/tomb
@@ -109,6 +109,8 @@ typeset -aH TOMBTMPFILES	  # Keep track of temporary files
 typeset -aH TOMBLOOPDEVS	  # Keep track of used loop devices
 typeset -A TOMBFILESSTAT      # Keep track of access date attributes
 
+typeset _MSG_FD_OVERRIDE # if set, _msg will write to this file descriptor
+
 # Make sure sbin is in PATH (man zshparam)
 path+=( /sbin /usr/sbin )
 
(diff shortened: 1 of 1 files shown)


jcrd:
I removed the last commit, but the same test failed. Upon investigation, I realized `gen_key` now prints messages to stdout, which will be written to the `TOMBKEYFILE` in `forge_key`. Decryption fails because the KDF header is missing from the first line.

jcrd:
All tests pass locally with the additional two commits, so I'm not sure what's happening...

jaromil:
It may be weirdness of Travis and its ever-changing Ubuntu install, I will also run tests whenever possible on our own CI at dyne, but will have to wait a bit for this.

jcrd:
Thanks for testing locally.

I rebased on master thinking the check from #332 might shed some light on the issue. As it turns out, the KDF test is passing, but [this](https://github.com/dyne/Tomb/blob/master/extras/test/30_kdf.sh#L12) test code is still [reporting](https://travis-ci.org/dyne/Tomb/builds/452103315#L1795) `bad key`.

I hadn't run the now failing steganography test locally since I was missing steghide. Oops.
The test expects the `exhume_key` function to print the key to stdout, but it now also prints all non-failure messages. Seems like my changes were more disruptive than anticipated.

One solution would be to pass the file descriptor as the first argument to `_msg` in the wrapper functions, so that a wrapper such as `_success2` could be created that prints in the success style but to stderr. These variants would be used in functions that rely on their stdout being captured. Any other ideas?

jaromil:
Thanks for following up! it looks good to me, would you mind rebasing to a single commit? then I would be happy to merge this.

jcrd:
Done. Glad we got this sorted out.

jaromil:
interesting about the echo speedup. I am using print since is a zsh builtin and supposedly faster than echo. wouldn't be better then to use `print ""`?

jaromil:
however, welcome among the Tomb contributors!

jcrd:
The speedup is in circumventing `gettext` which is called in `_msg`.
`type echo` says it's a builtin, but `print` would've also worked and might've looked more consistent alongside `_print`...

PR #332, opened 2018-11-03
Check for KDF header in is_valid_key

jcrd:
This check would've caught the invalid header being generated in #331.


jaromil:
gotcha! thanks.

Issue #306, opened 2018-02-02
Support pitchforkedsphinx for password handling


jaromil:
by stef, see https://github.com/stef/pitchforkedsphinx

heat-wave:
What is the scope of password handling expected here? Supporting 'get' operation will be relatively straightforward, while other operations can be more clunky. What I currently have in mind is adding **--sphinx-host** and **--sphinx-username** parameters to operations involving **ask_password** (an obvious example is **open**), and using pinentry to receive the master password for sphinx.

jaromil:
I think that's it, mostly. Perhaps we can also facilitate to put a key's password into sphinx?

heat-wave:
This can be done, although I believe it's important not to make this confusing to the user. There's two distinct passwords: _master_ (human-memorizable) and _generated_ (for lack of a better word; high-entropy, hard to remember) using the aforementioned master. Using the key's password as master serves no purpose.

We can try to make this seamless for the user (apart from configuring the oracle location for pwdsphinx): if they use the sphinx option, the password that we're asking when creating the tomb will serve as master for sphinx, and the result of the sphinx operation will serve as actual password of the key. Then unlocking follows the same logic and there's no confusion. Host and username parameters should still be left to the user (I would advise against auto-generating those on their behalf).

jaromil:
ACK. Some remarks
1. `master` : `generated` = `password` : `key` (euclidean comparison with Tomb's key system)
2.  the host-username pair shouldn't be used to generate secrets, since tombs are portable across systems, or is there no problem?
3. I don't think that a sphinx-made `generated` should be the `password` of a tomb key: I think that the tomb `key` is substituted with sphinx-made `generated` and that the `master` becomes the `password`.

What do you think? I think 3. simplifies a lot for the user, but I'm insecure about 2. (have to look more into sphinx, grateful for your insights)

heat-wave:
1. Noted.
2. Host and username are not system-specific, these are the values that allow the user to distinguish between various stored passwords (and since the storage itself isn't aware of its contents, this is necessary). Actual hostname and username of the user making the call don't matter here.
3. This does indeed remove a step from the process. AFAIK, the output format of sphinx is always a 40-character ASCII string. Do we then use it as TOMBSECRET to directly open/close tombs? Maybe I am too used to the standard keys and got confused here.

jaromil:
2. Noted :^)

we are left with the last point then:

[...]

PR #342, opened 2019-02-17
Cloakify integration

### LINKED ISSUE #286: support cloakify for key storage
Awesomeness beyond steganography: https://github.com/TryCatchHCF/Cloakify

heat-wave:
Fixes #286 
Summary:

- Cloakify added as a git submodule in /extras
- Steganographic functionality added as **tomb cloak/uncloak** commands rather than **bury/exhume** subcommands (since that was much cleaner code-wise)
- The user is expected to have **cloakify** and **decloakify** on their path, as well as python to run them
- Test suite updated
- Adding noise to resulting text is not supported (this can be done but makes usage less straightforward)

Most of the code is reusing parts of existing codebase. Please let me know if something is undone or can be improved.


heat-wave:
Travis build has passed, but it's worth noting that the cloakify-related tests did not run because of unsatisifed dependencies (hadn't thought of it prior to making the PR). Should I update the config to include them?

roddhjav:
Yes, you should ;)

jaromil:
Thanks for the contribution! lets make sure tests run and leave some time to read your changes. I am insecure about including cloakify as a submodule or jusy statically in Tomb. Would prefer the latter

heat-wave:
I wasn't sure whether to include raw sources or git submodule either. Given the license, I believe it can be copied and modified, so that isn't a problem. And since the easiest way to make the cloakify/decloakify scripts executable is to add a shebang at the start, if we are to choose the static sources, this simplifies testing.
If you're ok with updating the source in the unlikely event that more commits happen in the cloakify repo (or if we're happy with the scripts as they are), I'll make these changes shortly.

roddhjav:
> unlikely event that more commits happen
I would say, in the very likely event more commits happen. It not because is seems final now it will always be.

Cloakify is before all an optional dependency of Tomb, so why not simply add it to the dependency list and let the maintainer/user manage the installation separately. Furthermore, Cloakify has [releases](https://github.com/TryCatchHCF/Cloakify/releases), therefore, it makes no sense to include it with Tomb.

heat-wave:
Fair enough.

My problem is that Cloakify doesn't have an installation mechanism, so even if the user downloads it, there's quite a few assumptions to be made about how it's installed and used. No matter which way the dependency will be managed (I trust you the maintainers know better), I'd like some certainty on the above.

jaromil:
I have included all optional dependencies in Tomb releases, to avoid on-line connectivity reliance. But in case of python and other more complex setups this makes no point. There is also an archival value to it in case some software disappear (steghide being left unmaintained for long...) however I see all this doesn't apply to cloakify, so perhaps lets follow @roddhjav's suggestion and include only a wrapper to facilitate calling the system wide cloakify if necessary.

heat-wave:
Sure, I'll remove the submodule then. What kind of wrapper do you mean, though? If we assume that the user installs cloakify and therefore has the scripts on their path, it might be easier to just rewrite the calls to **python2 cloakify/decloakify** (and it would be consistent with the suggested usage in cloakify repo). Am I missing something?

heat-wave:
Well, after some struggle with paths, the new tests do run and pass. Apologies for messy commits.

jaromil:
Great! I'll read through the code ASAP. Meanwhile no worries for the messy commits, that's why PRs are there. Can you `git rebase -i HEAD~10` and squash your commits into one, with a clear comment? then push force to your branch and it will all be clean.

heat-wave:
Done. If/when we're finished with cloakify, I could look into integration with pitchforked-sphinx as per #306, if it is worth pursuing.

jaromil:
dear heat-wave, congratulations! this looks very neat. we know shell scripting is not rocket science, but it also takes a lot of love and passion to write something that is simple, readable and trustworthy. I think you have done very well here, so be welcome to the Tomb contributors. I'm also happy to read you like to do more :^) ciao

Issue #343, opened 2019-02-20
Error locking new container


tpoliaw:
I have used tomb in the past and it's always worked fine. Trying to get it to work on a new machine is giving errors.

Tomb version: 2.5
ZSH version: 5.7.1
OS: Arch (kernel 4.20.10-arch1-1-ARCH)
cryptsetup: 2.1.0

Following the example in the readme (dig/forge/lock) gets as far as the lock stage then fails with 

    tomb [W] cryptsetup luksFormat returned an error.
    tomb [E] Operation aborted.

I have made a container using cryptsetup directly and it seems to work ok. I can't try using the cryptsetup command in the readme because I don't have a container to try opening yet.

I tried to run using v2.4 (from the tag in git) and the current master version and both had the same behaviour.

The full trace is [here](https://gist.github.com/tpoliaw/8587f0805a9c7947dcd7a5e7c72b1265) if it helps. I am happy to get any other information if it helps.

Thanks for Tomb. It's a great script (when it's working).

jaromil:
Hi tpoliaw,
thanks for reporting, this sounds weird, arch has been prominent in packaging tomb since ages. Running with `-d` may give more info.

Narrat:
Cryptsetup was recently updated to 2.1
Without testing anything, just reading the changelog, I noticed one point which changed default behaviour. With 2.1 LUKS2 is used per default and maybe this causes an issue with newly created tombs?

Narrat:
Oh and header changes

jaromil:
In 2.0.6 manpage says
```
To use LUKS2, specify --type luks2.
```

Narrat:
Gosh. I overlooked one important word :D changes since 2.0.6 and not what I read : changes in 2.0.6 :D
So, yeah. Nothing to do with 2.0.6 (and it would have been weird to change something like that with just a Patchlevel increase)

roddhjav:
I actually have the same issue. I can confirm that adding `--type luks1` [line 1198](https://github.com/dyne/Tomb/blob/master/tomb#L1199) fix the issue. However, it still cause an issue with former version. Meaning, the regression tests for Tomb 2.3 do not work.

Furthermore, it would be nice to support Luks2.

tpoliaw:
> Running with `-d` may give more info.

The linked output includes the `-D` debug option (is that different to `-d`? It said that one wasn't recognised) for the lock command. I can include the debug output from the other two commands if that'd help.

[...]

Issue #262, opened 2017-05-24
Support for Docker


sanbor:
I'm always looking for a multi-platform encryption system. Docker now it's pretty easy to install in macOS and Windows 10. I'd like to hear what do you think about using docker to run tomb. I'm sure that there are security implications by doing this.

I successfully created and opened a tomb but I'm not sure how to close it, as the container exits after creating the tomb.

This might be an interesting approach to tackle multi-platform support.

jaromil:
interesting! never used Docker (I'm more of an LXC and KVM fundamentalist) but I do see the advantage to ease usage multiplatform and guess this can be the only way to export some of the integrity in computation that Linux dm-crypt offers into other environments that are fundamentally unsafe.

I'd say that the RNG quality is the major concern if using it mission critical, since the random generator in Win or OSX cannot be trusted at all. However there are ways to circumvent that with external RNGs.

All in all I think this is worth the effort and if you publish something I'll do my best to facilitate it into distribution. Happy hacking!

parazyd:
Keep in mind if you wish to use the tomb outside of the container, it might be a tough nut to crack, since both Windows and OS X are incompatible with ext{2,3,4} filesystem AFAIK.

jaromil:
If this thing goes forward I am willing to adapt tomb to work with FAT32. This may be a good occasion also to add support for other filesystems like btrfs.

roddhjav:
I can confirm tomb works in a docker container (I did a few tests for a Travis.yml that would run the tests suite on multiples distribution).

However, stop me if I'm wrong but dmcrypt is included in the Linux kernel, the container does not have a kernel and therefore all the call to dmcrypt are executed by the host. This is fine when the host is a Linux distribution, but I fail to see how this can work on windows.

reiven:
Keep in mind that to run a linux container on Windows, a linux virtual machine is installed, so in practice windows users will be running Tomb inside a VM.

jaromil:
Do docker images on Win or OSX manage the VM for the users transparently?
Is there some automatism to share a directory between the guest and host?

sanbor:
@jaromil there is a way of sharing directories between Docker and the host machine. It is called volumes.

Here is an example of how you could do it:
    docker run -it -v $PWD:/tomb ubuntu bash

That will create a new ubuntu container with the current host directory mounted in `/tomb`.

jaromil:
[...]

PR #344, opened 2019-03-14
Pitchforked sphinx integration

### LINKED ISSUE #306: Support pitchforkedsphinx for password handling

heat-wave:
Fixes #306 

**Summary**:

- updated **ask_password** and adjacent logic to support retrieving password via sphinx
- added sphinx opts to commands that expect password
- added sphinx installation script (also copies default sphinx/oracle config for testing)
- created base dockerfile
- added tests for commands using sphinx

Note for the users of this feature: sphinx (client) needs to be in quiet mode to work with tomb, verbose will disrupt picking up the password properly. As suggested in sphinx's description, **do not** run sphinx and oracle on the same machine (this only gives a false feeling of security). The configuration included in extras is for testing purposes only.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,26 +1,12 @@
-dist: trusty
 sudo: required
 language: c
 
-addons:
-  apt:
-    packages:
-      - zsh
-      - gnupg
-      - cryptsetup
-      - pinentry-curses
-      - gawk
-      - libgcrypt20-dev
-      - steghide
-      - qrencode
-      - e2fsprogs
-      - python2.7
+services:
+  - docker
 
 before_script:
-  - make --directory=extras/kdf-keys
-  - sudo make --directory=extras/kdf-keys install
-  - sudo ./extras/install_cloakify.sh
+  - docker build -t dyne/tomb .
+  - docker run -it --privileged dyne/tomb /bin/bash -c "oracle & make test"
 
 script:
-  - make --directory=extras/kdf-keys test
-  - make test
+  - docker run -it --privileged dyne/tomb /bin/bash -c "make --directory=extras/kdf-keys test"
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,14 @@
+FROM dyne/devuan:beowulf
+
+RUN apt-get update -y -q && apt-get install -y -q zsh cryptsetup gawk libgcrypt20-dev steghide qrencode python python2.7 python3-pip python3-dev libsodium-dev libssl-dev make gcc g++ sudo gettext file bsdmainutils
+RUN pip3 install setuptools wheel
+
+COPY . /Tomb/
+
+WORKDIR /Tomb/extras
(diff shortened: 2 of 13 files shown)


jaromil:
I've met @stef last weekend and noticed him about this :^)

stef:
wow. nice work. however there is not much gained if the sphinx oracle runs on the same host as the user inputs their password. if you run both the client and the server on the same host you essentially do this: `hash(password || hash(password)^seed)` which i believe is weaker than doing a proper argon2i on the password.

heat-wave:
Thank you @stef! I do understand that the oracle is meant to be on a different server, however for testing purposes it seemed easier to run both sphinx and oracle on the same box with tomb tests. I guess oracle could be extracted to a separate docker container and then the tests would be run by docker-compose (if jenkins supports that), but is it worth going to these lengths?

@jaromil if I may ask, does my docker config make sense (if we can complete it with loop device support)?

stef:
oh. if this is only for testing-purposes, it is of course totally ok to do so. but in a real-life setting the two components should be separated.

heat-wave:
Finally managed to run the tests using luks version fix (which I didn't include in the commit), summary:
- tests passing (apart from regression tests as mentioned in #343)
- changed sphinx.cfg to use fixed directory in /tmp
- updated cloakify instalalation script (since the wrapper was previously ignoring arguments)
- I haven't removed dockerfile yet; the progress of #262 looks promising so maybe I should wait until we have proper docker config and then update or reuse it

jaromil:
Thanks for the review @stef <3 and don't worry, Tomb is all about making it easier for people to keep keys and data on separate storage supports.

gregtzar:
@heat-wave Running a docker container as `PRIVILEGED` is one way to get loop devices working.

heat-wave:
Finally got to iron out some issues around using sphinx under docker. Current dockerfile yields a testable image (with **docker run -it --privileged _%imagename%_ /bin/bash -c "oracle & make test"**), the only problem left is that bind-hooks test fails under docker with error "tomb [W] bind-hook target not existent". I'll look into that when I can, but I got the same error when using @gregtzar 's image to run tests on a stable Tomb release.

heat-wave:
Alright, I got it all working. If this result is good enough, I'll pack things up into one commit and maybe tune dockerfile/travis config if you find any inconsistencies there.

jaromil:
Awesome work. Congrats for enduring until last mile. Excellent. Tested locally and works. Please pack and then will merge. Great motivation to fix last pending bugs and make a new release. Thanks!

heat-wave:
Thank you for the kind words! Genuinely happy to contribute.
I've packed the commits, seems ready to merge. I might not be as helpful in fixing 'real' bugs with encryption etc., but if more feature requests arise, I'll be on the lookout.

PR #350, opened 2019-05-22
switch default random source to /dev/urandom

### LINKED ISSUE #349: Can't forge a key
QQ截图20190513161327

It's too slow to generate

version: ubuntu 18  + tomb2.5

jaromil:
there is no need to default to a blocking source of random,
since /dev/random doesn't improves the quality of randomness in Linux

fix #349

### DIFF
--- a/doc/tomb.1
+++ b/doc/tomb.1
@@ -1,4 +1,4 @@
-.TH tomb 1 "April 16, 2017" "tomb"
+.TH tomb 1 "May 22, 2019" "tomb"
 
 .SH NAME
 Tomb \- the Crypto Undertaker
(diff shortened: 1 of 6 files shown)


doc/tomb.1
```diff
@@ -38,18 +38,18 @@ data gathered from a non-blocking source (/dev/urandom).
...
-encryption), then the \fI-r\fR option indicates the recipient key;
-more recipient GPG ids can be indicated (comma separated). The default
-cipher to protect the key is AES256, a custom one can be specified
-using the \fI-o\fR option, for a list of supported ciphers use
-\fI-v\fR. For additional protection against dictionary attacks on
-keys, the \fI--kdf\fR option can be used when forging a key, making
-sure that the \fItomb-kdb-pbkdf2\fR binaries in \fIextras/kdf\fR were
-compiled and installed on the system.
+random data from a non-blocking source (/dev/urandom) and it may take
+long only in some cases; to switch using a blocking source the
```
Narrat: Does it still takes a long time? My experience so far was, that only the usage of `/dev/random` could create those long times.

Narrat:
https://github.com/dyne/Tomb/blob/master/tomb#L1833
https://github.com/dyne/Tomb/blob/master/tomb#L1950
Maybe some additional places that should be adjusted?
Imo the warning can be removed or only displayed if `--use-random` is set and as the modus operandi changed, the comment itself should be adjusted?

PR #353, opened 2019-06-27
Ensure the gpg key is trusted.

roddhjav:
Hi,

This PR ensure GPG key is trusted. 

The field name from the `trusted` array are detailed in the [gpg doc](https://github.com/gpg/gnupg/blob/master/doc/DETAILS#field-2---validity).

In order to test this change, I needed to add a new untrusted GPG key to the test keyring. I also converted the keyring to the common keyring format because it is now supported everywhere and easier to manage.


jaromil:
Awesome!

Issue #164, opened 2014-11-22
_whoami function: not valid in this context: /root


fsLeg:
Just tried to test the latest snapshot of Tomb. I can create tombs normally, but whenever I need to enter a password for a key I get this error:

```
_whoami:local:30: not valid in this context: /root
tomb [W] User aborted password dialog.
tomb [E] No valid password supplied.
```

Latest stable (1.5.3) doesn't seem to have this function and I don't have any problems manipulating the keys or opening tombs with it. But I can't do that with the latest snapshot.

My system is Slackware64 14.1.

PR #359, opened 2019-08-19
Use getent instead of directly parsing passwd

jpducassou:
Hi. Please consider my patch.
When getting the home directory of the current user the file /etc/passwd was parsed instead of using getent like everywhere else.
When the current user is not local, for instance a ldap user, this causes the home directory to be guessed incorrectly. And because of that  bind-hooks execution is aborted.
Hope this helps.
Please tell if I have to make any adjustments.
Cheers!

### DIFF
--- a/tomb
+++ b/tomb
@@ -245,7 +245,7 @@ _whoami() {
 	# }
 
 	# Force HOME to _USER's HOME if necessary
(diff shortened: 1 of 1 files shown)


Narrat:
Currently checking if there is a reason, why the current approach was used.
Until now I stumbled upon https://github.com/dyne/Tomb/issues/164 . If it's made sure of, this cannot reappear?

PR #373, opened 2020-02-04
Fix manpage

dcommisso:
Hi,
I've fixed some stuff in linux manpage:
* replaced old "post-hooks" with the new "exec-hooks"
* Firefox example - Firefox doesn't start if specified profile directory doesn't exist


jaromil:
Thanks!

Issue #377, opened 2020-03-22
Failed to decrypt tomb on USB auto-mount


quickcougar:
Following the example from the documentation for decrypting a tomb containing a GnuPG directory, I would like to be able to:

- Store a GnuPG encrypted tomb key containing a GnuPG home on a USB flash drive
- Store a tomb containing a GnuPG home locally
- **Auto-mount** the USB flash drive and open the tomb
- Prompt the user (only once) for the password to unlock the forged key
- Use the bind-hooks tomb feature to map the GnuPG home directory to $HOME/.gnupg

Issue #140, appears to be similar in identifying the symptom, however, the solution does not appear to apply to auto-mounts.

```bash
Mar 22 14:24:43 hostname systemd[919]: media-gpg.mount: Succeeded.
Mar 22 14:24:48 hostname systemd[919]: Starting Decrypt gpghome on mount...
Mar 22 14:24:48 hostname open-tomb[70675]: 2020-03-22 14:24 USER: user
Mar 22 14:24:48 hostname open-tomb[70675]: 2020-03-22 14:24 LANG: en_US.UTF-8
Mar 22 14:24:48 hostname open-tomb[70675]: 2020-03-22 14:24 Checking if /dev/sdc1 is mounted ...
Mar 22 14:24:48 hostname open-tomb[70675]: 2020-03-22 14:24 DISPLAY: :0
Mar 22 14:24:48 hostname open-tomb[70675]: 2020-03-22 14:24 tty: not a tty
Mar 22 14:24:48 hostname open-tomb[70675]: 2020-03-22 14:24 TTY: /dev/pts/2
Mar 22 14:24:48 hostname open-tomb[70675]: 2020-03-22 14:24 GPG_TTY:
Mar 22 14:24:48 hostname open-tomb[70675]: 2020-03-22 14:24 GPG_AGENT_INFO:
Mar 22 14:24:48 hostname open-tomb[70675]: 2020-03-22 14:24 Device mounted at /media/USB-FLASH-DRIVE
Mar 22 14:24:48 hostname open-tomb[70675]: 2020-03-22 14:24 Opening tomb /home/user/gpg.tomb ...
Mar 22 14:24:48 hostname open-tomb[70675]: 2020-03-22 14:24 /usr/bin/tomb --version
Mar 22 14:24:48 hostname open-tomb[70675]: 2020-03-22 14:24 tomb version:   Tomb 2.7 - a strong and gentle undertaker for your secrets
[...]

PR #378, opened 2020-03-23
Allow pinentry to fail, so long as the password comes through.

quickcougar:
Fix for issue #377


jaromil:
ACK. Thanks again, this is useful. Even more useful would be to change the ERR to a WARN in pinentry, but I'm not sure about following up on it.

Issue #385, opened 2020-11-03
[security] Any password unlocks tomb when using pinentry-curses and non-empty $DISPLAY


aaronjanse:
Thank you for this tool. I've been using it for a while, and I appreciate its straightforwardness.

I recently noticed that my tomb unlocks no matter what password is provided.

Steps to reproduce:
1. Use `tomb` with pinentry-curses in the $PATH
2. Run `tomb dig -s 100 foobar`
3. Provide any password for `tomb forge foobar.key`
4. Provide any password for `tomb lock foobar -k foobar.key`
5. Provide any password for `tomb open foobar -k foobar.key` (!!!)

My operating system is NixOS.

jaromil:
@aaronjanse weird, can you give us the output of tomb -v

aaronjanse:
```
$ tomb -v
  Tomb 2.7 - a strong and gentle undertaker for your secrets

   Copyright (C) 2007-2017 Dyne.org Foundation, License GNU GPL v3+
   This is free software: you are free to change and redistribute it
   For the latest sourcecode go to <http://dyne.org/software/tomb>

   This source code is distributed in the hope that it will be useful,
   but WITHOUT ANY WARRANTY; without even the implied warranty of
   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
   When in need please refer to <http://dyne.org/support>.

  System utils:

gpg: WARNING: unsafe permissions on homedir '/home/ajanse/.gnupg'
gpg: WARNING: unsafe permissions on homedir '/home/ajanse/.gnupg'
  Sudo version 1.9.3p1
  cryptsetup 2.3.4
  pinentry-curses (pinentry) 1.1.0
  findmnt from util-linux 2.36
  gpg (GnuPG) 2.2.23 - key forging algorithms (GnuPG symmetric ciphers):
  IDEA 3DES CAST5 BLOWFISH AES AES192 AES256 TWOFISH CAMELLIA128 CAMELLIA192 CAMELLIA256

  Optional utils:

  /nix/store/vkmnfr10cbz8799yvwy352607jrww92h-gettext-0.21/bin/gettext
  dcfldd not found
  /run/current-system/sw/bin/shred
  steghide not found
  /nix/store/zzsycw09gljk0ay6wsnz70vf47dp9z83-e2fsprogs-1.45.5-bin/bin/resize2fs
  tomb-kdb-pbkdf2 not found
  qrencode not found
  swish-e not found
  unoconv not found
[...]

PR #386, opened 2020-11-10
Use _verbose for pinentry-curses in ask_password

aaronjanse:
Fixes https://github.com/dyne/Tomb/issues/385. Note that affected users will have to decrypt their existing tombs using the password `tomb [W] Detected DISPLAY, but only pinentry-curses is found.`.

Should I append KNOWN_BUGS in this PR, or would a maintainer like to write that part?


jaromil:
@aaronjanse I'll write that part upon release (which will come very soon now) thanks!

Issue #387, opened 2020-11-13
Avoid mounting same tomb twice on a user specified directory


shaform:
Tomb 2.7 is currently allowing users to mount the same tomb multiple times if users provide a target directory:

```
$ tomb dig -s 100 secret.tomb
$ tomb forge secret.tomb.key
$ tomb lock secret.tomb -k secret.tomb.key
$ mkdir -p target_dir
$ tomb open secret.tomb target_dir -k secret.tomb.key
$ tomb open secret.tomb target_dir -k secret.tomb.key
$ tomb open secret.tomb target_dir -k secret.tomb.key
```

However, doing so result in lost of data as mentioned in roddhjav/pass-tomb#27 or #326.

It would be preferable to avoid this behavior.

jaromil:
Thanks for the reminder, you are welcome to review the PR #389

Issue #326, opened 2018-10-05
Avoid mounting same tomb twice or on already used mountpoints


stevelitt:
When the user makes the mistake of running tomb open multiple times,  it is allowed. Each extra open performs the mount on top of the last mount point, masking it. Each extra open creates an extra /dev/mapper/tombname.tomb.number.loopDigit. tomb list lists all of them, but on all but one says they're not available because they're masked by another tomb.

There's no pure Tomb way out of this. You can repeatedly perform sudo tomb close tombname until it reports no tombs open, but /dev/mapper still contains the extra loops. Those are gotten rid of only by repeatedly performing cryptsetup close tombname.tomb.*

Would it be better if Tomb immediately check the eventual mountpoint pathname for existance, and refuse to act if it is: Hopefully with an informative error message.

This isn't a bug, because the cause is user error. But the user making the error is presented with a very difficult situation, so maybe a barrier should be erected to prevent it.

jaromil:
Good point, some time in the past I thought we fixed this (refusing to mount the same tomb twice) but perhaps I've overlooked it. Should be an easy fix.

amerlyq:
Problem still exists and it's not minor but major one :(
Look here for bug description and consequences:
https://github.com/roddhjav/pass-tomb/issues/27
It's a pity, but I couldn't debug the problem root cause by my own limited ability to provide a fix.
P.S. I will copy issue description here too if you request it.

PR #389, opened 2020-11-13
improve the check if a tomb file is in use

jaromil:
Change the mapper path using a hash of the tomb file path,
making it unique and reproducible to check if tomb is in use.
Check happens inside the new render_mapper() function which is
executed right after the key file opening.

Provides a fix for #387 and #326 as well signaled in https://github.com/roddhjav/pass-tomb/issues/27

### DIFF
--- a/tomb
+++ b/tomb
@@ -647,6 +647,26 @@ is_valid_tomb() {
 	return 0
 }
 
+# render the path to the unique /dev/mapper using an hash of the path
+# of the tombfile and its name. Checks for duplicates (tomb is in use)
+render_mapper() {
+	[[ "$tombpath" == "" ]] &&
+		_failure "cannot render mapper: missing \$tombpath"
+	[[ "$TOMBNAME" == "" ]] &&
+		_failure "cannot render mapper: missing \$TOMBNAME"
+	local maphash=`print $tombpath | sha256sum -z`
+	mapper="tomb.$TOMBNAME.${maphash[(w)1]}"
+	local mounted_tombs=(`list_tomb_mounts`)
+	local usedmapper
+	for t in ${mounted_tombs}; do
+		usedmapper=`basename "${t[(ws:;:)1]}"`
+		[[ "$usedmapper" == "$mapper" ]] &&
+			_failure "Tomb file already in use: ::1 tombname::" $TOMBNAME 
+	done
+	_verbose "Mapper: ::1 mapper::" $mapper
+	print "$mapper"
+}
+
 # $1 is the tomb file to be lomounted
 lo_mount() {
 	tpath="$1"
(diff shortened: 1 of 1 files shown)


shaform:
Thanks for the update. It does avoid the problem in roddhjav/pass-tomb/issues/27.

I've noticed a small problem though. Because the hash is computed based on the input path. If the users use the following commands, they could still mount the same file multiple times. But I guess this situation might be rare.
```
$ cd /tmp
$ tomb dig -s 100 secret.tomb
$ tomb forge secret.tomb.key
$ tomb lock secret.tomb -k secret.tomb.key
$ mkdir -p target_dir
$ tomb open secret.tomb target_dir -k secret.tomb.key
$ tomb open ../tmp/secret.tomb target_dir -k secret.tomb.key
$ tomb open ../tmp/../tmp/secret.tomb target_dir -k secret.tomb.key
```

shaform:
I also notice that if we use the version in this PR. Every time the tomb is closed. A new entry of a encrypted disk would be added to the sidebar of nautilus:
side bar

However, this issue does not exist in Tomb 2.7.

Not sure if this would have any negative impact except for cluttering the sidebar.

jaromil:
> I've noticed a small problem though. Because the hash is computed based on the input path. If the users use the following commands, they could still mount the same file multiple times. But I guess this situation might be rare.

nevertheless, good to have it fixed, thanks! using `realpath` now to computer always an absolute path.

jaromil:
> I also notice that if we use the version in this PR. Every time the tomb is closed. A new entry of a encrypted disk would be added to the sidebar of nautilus:
> However, this issue does not exist in Tomb 2.7.
> 
> Not sure if this would have any negative impact except for cluttering the sidebar.

I don't know Nautilus well and puzzled by the fact this wasn't happening in Tomb 2.7 already since the mountpoint haven't changed, just the length of the `/dev/mapper` file... would be good to receive advice from someone more expert about Nautilus or at least find its logic documented somewhere, freedesktop "standards" or the likes.

I'll leave this to another PR if any in the future that can improve desktop integration.

roddhjav:
I have the same issue. It seems the loop devices are not correctly detached (and still visible with `losetup`), it explains why nautilus shows them. Doing `losetup --detach-all` fix it but is not really a sustainable solution.

shaform:
I have done some tests and it seems that the problem arises when `.$(basename $nstloop)` is removed from the filename of the mapper.

jaromil:
My wild guess on a (undocumented feature?) in Nautilus: mounted `/dev/mapper` volumes with `.loop` extension don't get listed in sidebar? is there anything like this?

jaromil:
I've found out that is possible to hide mount drives from this setting:
`gsettings set org.gnome.shell.extensions.dash-to-dock show-mounts false`
which I believe may be the desired option for deniability, maybe worth mentioning in docs.

shaform:
@jaromil 

> My wild guess on a (undocumented feature?) in Nautilus: mounted `/dev/mapper` volumes with `.loop` extension don't get listed in sidebar? is there anything like this?

No, simply adding `.loop` does not work. It has to be `.loop[N]` and `[N]` has to be the actual number that is used for mapping the device.

As mentioned by @roddhjav, this probably is not specific to Nautilus, but rather, when the file name does not contain the actual loop device being used, the loop device would not be detached.

jaromil:
I have now seen the issue with the missing detachment of the loop file, there is a problem in carrying the state of the loop number and there are solutions I'll be experimenting with.

jaromil:
Please if you can share some feedback on #391 seems to fix it see https://travis-ci.org/github/dyne/Tomb/builds/744123599 also it improves the code by throwing away some and making some cleanup (had it in mind, long due)

PR #391, opened 2020-11-16
refactor of state tracking for loop mounting

jaromil:
testing a small refactor

simplified function calls for tracking of loop mount by using global
variables whose scope is limited to execution, most computation is now
included in the `is_valid_tomb` function.

### DIFF
--- a/tomb
+++ b/tomb
@@ -257,33 +257,6 @@ _whoami() {
 
 }
 
-# Define sepulture's plot (setup tomb-related arguments)
-# Synopsis: _plot /path/to/the.tomb
-# Set TOMB{PATH,DIR,FILE,NAME}
-_plot() {
-
-	# We set global variables
-	typeset -g TOMBPATH TOMBDIR TOMBFILE TOMBNAME
-
-	TOMBPATH="$1"
-
-	TOMBDIR=$(dirname $TOMBPATH)
-
-	TOMBFILE=$(basename $TOMBPATH)
-
-	# The tomb name is TOMBFILE without an extension and underscores instead of spaces (for mount and cryptsetup)
-	# It can start with dots: ..foo bar baz.tomb -> ..foo_bar_baz
-	TOMBNAME=${${TOMBFILE// /_}%.*}
-	# use the entire filename if the previous transformation returns
-	# an empty string. This handles the corner case of tomb being
-	# hidden files (starting with a dot) and have no extension (only
-	# one dot in string)
-	TOMBNAME=${TOMBNAME:-${TOMBFILE}}
-	[[ "$TOMBNAME" = "" ]] &&
-		_failure "Tomb won't work without a TOMBNAME."
-
-}
-
 # Provide a random filename in shared memory
 _tmp_create() {
 	[[ -d "$TMPPREFIX" ]] || {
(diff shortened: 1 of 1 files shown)


Narrat:
Didn't look at the changes, but it kinda looks like a permission problem is still there
If there wasn't any loop device initialized before, it will try to do such a thing and fail because of privileges.

```
$ tomb open test -k lalala.key                                                         
(...)
losetup: unused loop-device couldn't be found: unsufficient permissions
basename: missing Operand
„basename --help“ for more Information

[sudo] Enter password for Superuser-Rights
```
and on close:
```
$ tomb close                                                                                             
tomb  .  Tomb [test], mounted at /run/media/omnius/test, is getting closed
losetup: /dev/: use of the device failed: no such device found
tomb (*) Tomb [test] closed: (...)
```
And it didn't get closed completly. Like the reported issue with Nautilus: The mapper or loop still remained.
Curiously next time everything worked fine. I suppose because of the existing loop device.

tomb
```diff
@@ -597,67 +569,60 @@ is_valid_tomb() {
...
-render_mapper() {
-	[[ "$tombpath" == "" ]] &&
-		_failure "cannot render mapper: missing \$tombpath"
-	[[ "$TOMBNAME" == "" ]] &&
-		_failure "cannot render mapper: missing \$TOMBNAME"
-	local maphash=`realpath $tombpath | sha256sum -z`
-	mapper="tomb.$TOMBNAME.${maphash[(w)1]}.loop"
+	# checks if Tomb already mounted (or we cannot alter it)
+	local maphash=`realpath $TOMBPATH | sha256sum -z`
+	local nextloop=`losetup -f`
```
shaform, line 602: Right, it seems that if the loop device is not already present in `/dev/` then this line would cause a permission error.

jaromil: oddly enough tests have passed without problems

Narrat: But there the "error" also appears
```
(...)
tomb  .  Testing creation: lock

tomb [D] Identified caller: omnius (1000:100)
tomb [D] Tomb command: lock /tmp/test.tomb
tomb [D] Caller: uid[1000], gid[100], tty[/dev/pts/3].
tomb [D] Temporary directory: /tmp/zsh
tomb [D] is_valid_tomb /tmp/test.tomb
tomb [D] tomb file is readable
tomb [D] tomb file is a regular file
tomb [D] tomb file is not empty
tomb [W] File is not yet a tomb: /tmp/test.tomb
losetup: cannot find an unused loop device: Permission denied
basename: missing operand
Try 'basename --help' for more information.
tomb [D] Mapper: tomb.test.10abda9a4ec9d0e665fe25155685f2e9ddf8f98034aa1b35a8b3748099751b24.
tomb [D] tomb file is not currently in use
tomb  .  Valid tomb file found: /tmp/test.tomb
tomb  .  Commanded to lock tomb test.tomb
tomb [D] Tomb found: /tmp/test.tomb
tomb [D] Loop mounted on /dev/loop0
tomb [D] load_key argument: /tmp/test.tomb.key
tomb [D] load_key: /tmp/test.tomb.key
tomb [D] is_valid_key
(...)
```
In the end it passed the test, but didn't work correctly

PR #402, opened 2020-12-22
Add some French translation

deepin-mozart:
Add some French translation


jaromil:
thanks! one thing please note 'pinentry-curses' should be left as-is and not translated as les maledictions of pinentry, though considering the history of issues raised by pinentry it may be a fit translation 😂

PR #403, opened 2020-12-24
Read CFLAGS in kdf-keys makefile

catleeball:
Added `$(CFLAGS)` environment variable in the kdf-keys makefile to allow users to specify additional build flags.


jaromil:
useful, thanks!

PR #406, opened 2021-01-01
dig/forge unecessary sudo

mcrapet:
Minor fixes for avoiding  sudo chown

`tomb dig` and `tomb forge` don't need to be executed with sudo.

### DIFF
--- a/tomb
+++ b/tomb
@@ -1955,7 +1955,7 @@ forge_key() {
 		_failure "Operation aborted."
 	}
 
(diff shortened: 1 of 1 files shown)


mcrapet:
`sudo su -` weird case not tested.

jaromil:
Hi Matthieu! Looks like a nice cleanup and improvement, also useful to limit the need for suid operations. Thanks for this contribution, awaiting other reviews on anything I may have missed.

jaromil:
thanks again and welcome among Tomb's contributors!

PR #408, opened 2021-01-23
Fix configs and dockerfile to enable sphinx in tests

heat-wave:
- Installing **pwdsphinx** from pip3 still results in incorrect entrypoint, so now we're getting both **libsphinx** and **pwdsphinx** from the same place (more consistent)
- **sphinx** no longer _requires_ SSL cert, but fails without it (`check_hostname requires server_hostname`, seems like check_hostname in SSL context [stays True](https://github.com/stef/pwdsphinx/blob/0b2ae15ada58010469d9cb91c43aebf3ba9a893b/pwdsphinx/sphinx.py#L64) in that case); **oracle** requires it anyway so I kept the dummy cert creation in the sphinx installation script
- added chimaera repo to get the required version of libsodium (as I mentioned, compiling it from source takes a while)
- updated expected text for when sphinx command fails (it is no longer `ValueError: fail`)

### DIFF
--- a/extras/install_sphinx.sh
+++ b/extras/install_sphinx.sh
@@ -5,8 +5,11 @@ cd libsphinx
 git submodule update --init --recursive --remote
 cd src
 sed -i 's|/usr/local|/usr|' makefile
-make
-sudo make install
-ldconfig
-pip3 install pwdsphinx
-sudo mkdir -p /etc/sphinx
+make && make install && ldconfig
+cd ../..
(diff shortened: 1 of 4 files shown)


stef:
anything i can fix for you upstream?

jaromil:
Amazing work fixing this, also backporting libsodium libraries to beowulf from chimaera.

Many thanks

Issue #410, opened 2021-02-07
support doas


marcusatiliusregulus:
please support alternatives to `sudo`, such as `doas`. you could provide a flag `--sudo <users-choice>` like `yay` does...

jaromil:
@marcusatiliusregulus I wholeheartedly agree with this request, ready to accept PR proposals, else I will try to find the time to make one myself, that may take a month or two

heat-wave:
Added **doas** support in #412

marcusatiliusregulus:
thanks

PR #412, opened 2021-04-07
Add support for sudo alternatives (e.g. doas)

heat-wave:
- Replaced a stray unwrapped **sudo** usage with **_sudo** call
- **doas** does not support custom prompts, so the formatted message still only applies to **sudo**
- Not entirely sure how many tests we want to cover that functionality with; I added basic ones but can duplicate others (with the same **--sudo doas**) as well


extras/test/00_create.sh
```diff
@@ -18,13 +18,25 @@ test_expect_success 'Testing tomb creation: dig, forge and lock' '
...
-        | hexdump -C &&
-    tt_lock --tomb-pwd $DUMMYPASS --sphx-user $DUMMYUSER --sphx-host $DUMMYHOST
-    '
+        tt_dig -s 20 &&
+        tt_forge --tomb-pwd $DUMMYPASS --sphx-user $DUMMYUSER --sphx-host $DUMMYHOST &&
+        print $(echo $DUMMYPASS | sphinx get $DUMMYUSER $DUMMYHOST) \
+            | gpg --batch --passphrase-fd 0 --no-tty --no-options -d $tomb_key \
+            | hexdump -C &&
+        tt_lock --tomb-pwd $DUMMYPASS --sphx-user $DUMMYUSER --sphx-host $DUMMYHOST
+        '
```
Narrat, line 27: Also wrong indention. Seems that this should be a real tab.

tomb
```diff
@@ -110,16 +110,20 @@ export TEXTDOMAIN=tomb
...
 
-# Wrap sudo with a more visible message
+# Wrap sudo with a more visible message or apply user-supplied alternative to sudo
 _sudo() {
-	local msg="[sudo] Enter password for user ::1 user:: to gain superuser privileges"
-	command -v gettext 1>/dev/null 2>/dev/null && msg="$(gettext -s "$msg")"
-	msg=${(S)msg//::1*::/$USER}
-	sudo -p "
+	if option_is_set --sudo; then
+		$(option_value --sudo) ${@};
```
jaromil, line 116: Better parse only valid values here in a `case/esac` list to avoid execution injections of sorts.

tomb
```diff
@@ -287,7 +291,7 @@ _is_encrypted_block() {
 	lsblk --help | grep -Fq -- --inverse
 	[[ $? -eq 0 ]] && s="--inverse"
 
-	sudo lsblk $s -o type -n $b 2>/dev/null \
+	_sudo lsblk $s -o type -n $b 2>/dev/null \
```
jaromil, line 294: well spotted. thanks!

jaromil (changes requested):

tomb
heat-wave: Will do, but if I'm not mistaken, nothing is stopping an attacker from calling their script **doas** and doing something malicious under the hood. So this may lead to a false sense of security.

jaromil: It is just one small step towards better security. There are solutions for the issue you raise but they are beyond the scope of this change I guess.

jaromil (approved):

jaromil:
Many thanks for this contribution!!

marcusatiliusregulus:
as this is now merged ill close the doas support issue #410 

great work guys, thanks

PR #437, opened 2022-03-07
lo_mount: check for loop support needs privilege escalation

### LINKED ISSUE #436: losetup: cannot find an unused loop device
Hi!

I had tomb working but after an update I am getting the error:

```
losetup: cannot find an unused loop device: Permission denied
tomb [W] Loop mount of volumes is not possible on this machine, this error
tomb [W] often occurs on VPS and kernels that don't provide the loop module.
tomb [W] It is impossible to use Tomb on this machine under these conditions.
tomb [E] Operation aborted.
```

After downgrading it is working again.
Using `tomb-git` from AUR on Archlinux.

Thanks

Narrat:
If there is no free loop device, the call of loopsetup -f will create one and return it. For this it needs privilege escalation.
It doesn't need those, if there is already an used device, but that cannot be guaranteed.

Closes #436
________________
Just readds the logic before the change. Maybe there is a better way to check for loop support?

### DIFF
--- a/tomb
+++ b/tomb
@@ -630,7 +630,7 @@ lo_mount() {
 	tpath="$1"
 
 	# check if we have support for loop mounting
-	TOMBLOOP=`losetup -f`
+	TOMBLOOP=`_sudo losetup -f`
 	[[ $? = 0 ]] || {
 		_warning "Loop mount of volumes is not possible on this machine, this error"
 		_warning "often occurs on VPS and kernels that don't provide the loop module."


Narrat:
Looking at the whole `lo_mount()` function. Isn't there a chance for a race condition?
It checks for loop support and sets `TOMBLOOP`. If that succeeds it will mount the loop device with the next free device.
But in the small time window between the two commands something different could grab hold of the one saved in `TOMBLOOP` and the actual tomb would be mounted in a different device. But the incorrect one would be saved in `TOMBLOOPDEVS`.
Probably not likely to happen, but nevertheless sth that should be avoided at all?

jaromil:
Yes, there can be a race condition as you mention.

PR #443, opened 2022-07-20
Update Dockerfile

jensrischbieth:
Building the image fails as pinentry doesn't exist in the repository, only pinentry-curses. Image building works w/ only pinentry-curses.


jaromil:
Cheers. Beware that using tomb througu Docker is very, very experimental and we cannot help much troubleshooting due to excessive complexity and points of failure.

PR #450, opened 2022-10-25
cmd lock: extend parameters to --filesystem

chri2:
The --filesystem option can be used  to  specify
an  alternative  filesystem used to format the tomb, in place of the default "ext4".

Beside "btrfs" now the following parameters to --filesystem are supported:

"ext3"    using operating system defaults
"ext4"    using operating system defaults
"btrfs"   for tombs >= 47MB using operating system defaults
"btrfsmixedmode"    for tombs >=18MB btrfs mixed mode (see mkfs.btrfs(8))
"ext3maxinodes"     ext3 with a maximum of inodes (for many small files)
"ext4maxinodes"     ext4 with a maximum of inodes (for many small files)

These changes help use scenarios in which there is a great number of small files and/or directories in a small filesystem, like e.g. the pass-tomb extension to pass.

### DIFF
--- a/doc/tomb.1
+++ b/doc/tomb.1
@@ -66,10 +66,21 @@ option can be used to specify the cipher specification: default is
 If you are looking for something exotic, also try
 "serpent-xts-plain64". More options may be found in cryptsetup(8) and
 Linux documentation. The \fI--filesystem\fR option can be used to
-specify "btrfs" as an alternative filesystem used to format the tomb,
+specify an alternative filesystem used to format the tomb,
 in place of the default "ext4". This operation requires root
-privileges to loopback mount, format the tomb (using LUKS and Ext4),
-then set the key in its first LUKS slot.
+privileges to loopback mount, format the tomb (using LUKS and mkfs),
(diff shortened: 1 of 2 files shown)


jaromil:
very nice to have sane defaults at hand for common typologies of filesystems. Thanks again for this contribution!

chri2:
If I'd want to check from pass-tomb extension whether the new filesystem parameters are available, could I rely on the version being higher than the first release containing this MR or should we extend the `--filesystem` with some parameter (or no parameter at all) to output a list of supported parameters?

jaromil:
@chri2 well, I can tag a new release perhaps it is easier. Else I'd rather refer to the installed manpage (which I saw you updated, well done!)

chri2:
> Else I'd rather refer to the installed manpage

I was thinking about on how to find out from the pass-tomb code which filesystem options are available for the initial creating of a password storage. I'd like to see the default for the filesystem changing depending on the available version of tomb.

Short: A release would be great to have something to check for from pass-tomb :)

PR #465, opened 2023-01-15
extras/gtomb severe update.

nerun:
Essentially:
 - Removed all those calls to sudo password.
 - A few improvements to resolve questions regarding swapoff/swapon or flag force (-f), that now can be setup inside file. Sudo will be called just when user setup `SWAPOFF="true"`. By default `SWAPOFF="false"` what means that flag `-f` will be used.
- Also, i have fixed a small error in KDF question: when user chooses not to use KDF, window closes and key is not created, while in true it should be created but without KDF.

### DIFF
--- a/extras/gtomb/gtomb
+++ b/extras/gtomb/gtomb
@@ -18,15 +18,18 @@
 # You should have received a copy of the GNU General Public License
 # along with this source code. If not, see <http://www.gnu.org/licenses/>.
 
-ver="0.8"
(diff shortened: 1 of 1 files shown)


jaromil:
oh! hi! this extras script was long neglected, I can well believe there were things to improve. Also I like to see code being deleted and overall made simplier. It is not covered by tests so let's wait a bit more before merging. I am not a desktop user but will try it to see how it works now.

jaromil:
I have read the code and tried this on a test installation running budgie desktop, works well so I'll merge. Thanks for the contribution!!

PR #469, opened 2023-04-27
test macos with fuse-t

jaromil:


jaromil:
Further testing on a local machine lead to a lot of doubts... this may be possible to do, but will never be stable enough to endure the challenge of time and future OSX releases...

PR #475, opened 2023-06-23
mount_tomb: make use of ACL in specific locations

### LINKED ISSUE #461: mountpoint `/run/media/$USER` always with root ownership
After reboot, running `tomb open example.tomb -k example.tomb.key`, it mounts on  `/run/media/$USER` which with root ownership. As a normal user, somehow I cannot use the bash auto-completion to `cd` into the `/run/media/$USER/example`, I have to type the absolute path to `cd` into that. So every time I just use `sudo chown $USER:$USER /run/media/$USER` to change the ownership of that mountpoint (I do not know if it impact the security) so I can use the bash auto-completion. However after a reboot, this problem happen again. Is it written by design for security? Thank.

Narrat:
Namely /run/media/$USER.
The mount point itself is owned by root, therefore one needs to know the name of the mountpoint to change to the location. Other tools for mounting media like udisksctl set ACL to allow the owner to use it normally (autocompletion and such).

Fixes #461
_______________
To get the solution for the named issue going.
There were some undecided points and maybe with some changed code this can go on further.

Therefore this PR is only a draft. Request for comments

### DIFF
--- a/tomb
+++ b/tomb
@@ -65,6 +65,7 @@ typeset -i RESIZER=1
 typeset -i RECOLL=1
 typeset -i QRENCODE=1
 typeset -i LSOF=1
+typeset -i ACL=1
 
 # Default mount options
 typeset		 MOUNTOPTS="rw,noatime,nodev"
@@ -897,7 +898,7 @@ function _print() {
 _list_optional_tools() {
 	typeset -a _deps
 	_deps=(gettext dcfldd shred steghide)
-	_deps+=(resize2fs tomb-kdb-pbkdf2 argon2 qrencode recoll unoconv lsof)
+	_deps+=(resize2fs tomb-kdb-pbkdf2 argon2 qrencode recoll unoconv lsof setfacl)
 	for d in $_deps; do
 		_print "`which $d`"
 	done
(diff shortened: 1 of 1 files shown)


jaromil:
This is especially interesting if it allows us to remove the `chown` to $USER in case ACL==1 and just use setfacl.

Narrat:
Interesting point. Theoretically it would allow that. But then it probably needs a better check for ACL support. Let's see what can be done.

jaromil:
We need more exhaustive testing for this, many corner cases can arise... Should we leave the code as-is? Perhaps ACL as an opt-in rather than auto-detect.

Narrat:
This is definitively nothing for 2.10 :D

I lean to split the approach:
* Inside of the tomb
* the eventual mointpoint

For the first point we can use auto-detect or entirely switch, as there is control over the initial tomb. But could also be be opt-in. Set a flag and a ACL config file is placed beside the other tomb files in a tomb. If it's there apply that instead of the chown.
Regarding the second point.. yeah... Opt-in may be the way to go. But still not really sure how to approach it the best (although I must admit this got again lost in daily chores...)

Narrat:
The scope of this PR adjusted to its original intent to match the behaviour when mounting a tomb at `/run/media/$USER/`.
Additional thoughts on replacing `chown` or support `ACL`-files in a tomb aren't forgotten, but should discussed separately. I got some ideas on how to proceed with those cases, but currently there are enough open construction sites :D

Narrat:
Rebasing seems to be have something which it shouldn't have :D It reintroduced partly old behaviour of `tomb slam`
```
[[ $LSOF == 0 ]] && {
					unset SLAM
					_warning "lsof not installed: cannot slam tombs."
					_warning "Trying a regular close." }}
```
https://github.com/dyne/tomb/blob/master/tomb#L3321-L3325
is now back.

PR #476, opened 2023-06-23
Remove unused check on python2

Narrat:
As far as I can see nothing in `tomb` directly requires `python2`.
And the check itself wasn't used anywhere?

Unless it is parsed from somewhere else? Related to tests?

If not, why not getting rid of it?


jaromil (approved):
makes sense to remove it, must be a leftover. Thanks!

PR #479, opened 2023-06-25
Manpage update

nerun:
Fixed wrong information about `--kdfmem`.

```
2^18 = 262,144 kilobytes = 250 MiB
```
Not 262 MiB as stated before.


jaromil:
good catch

Issue #464, opened 2023-01-15
tomb index / search and mlocate


nerun:
Please, remove the need to just `mlocate` to be installed, instead add some alternatives like `locate`, `plocate` etc.
```
$ tomb index
tomb [E] Index command needs 'mlocate' to be installed.
$ tomb search "test"
tomb [E] Index command needs 'mlocate' to be installed.
```

In Linux Mint, `mlocate` is just a dummy package. I have installed but when use mlocate in terminal, nothing happens:
```
$ mlocate
zsh: command not found: mlocate
```
Even if i add a script pointing to locate, it doesn't works:
```shell
#!/usr/bin/env zsh
locate "$@"
```
Would be better if `tomb` consider some alternatives.

nerun:
```
$ tomb --version
  Tomb 2.9.0 - a strong and gentle undertaker for your secrets

   Copyright (C) 2007-2021 Dyne.org Foundation, License GNU GPL v3+
   This is free software: you are free to change and redistribute it
   For the latest sourcecode go to <http://dyne.org/software/tomb>

   This source code is distributed in the hope that it will be useful,
   but WITHOUT ANY WARRANTY; without even the implied warranty of
   MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
   When in need please refer to <http://dyne.org/support>.

  System utils:

  zsh 5.8.1 (x86_64-ubuntu-linux-gnu)
  Sudo versão 1.9.9
  cryptsetup 2.4.3
  pinentry-gnome3 (pinentry) 1.1.1
  findmnt de util-linux 2.37.2
  gpg (GnuPG) 2.2.27 - key forging algorithms (GnuPG symmetric ciphers):
  IDEA 3DES CAST5 BLOWFISH AES AES192 AES256 TWOFISH CAMELLIA128 CAMELLIA192 CAMELLIA256

  Optional utils:

  /usr/bin/gettext
  dcfldd not found
  /usr/bin/shred
  /usr/bin/steghide
  /usr/sbin/resize2fs
  /usr/libexec/tomb/tomb-kdb-pbkdf2
  /usr/bin/qrencode
  swish-e not found
  unoconv not found
  /usr/bin/lsof
```

nerun:
Fixed in PR #483

Issue #482, opened 2023-06-26
Translations help - perl or shell script?


nerun:
I have started a translation to Brazilian Portuguese (pt_BR), it's a WIP, but 44% complete.

I noticed that `extras/translations/tomb.pot` was outdated so I ran the perl script `generate_translatable_strings.pl` (just because Makefile uses it) to create a new version and then created a `pt_BR.po` file.

I have tried shell script `generate_translatable_strings.sh` but it has some issues and duplicated strings. Poedit can remove duplications anyway.

Since i have started with the old tomb.pot i ran `msgmerge --update pt_BR.po tomb.pot`

My question is: to create a new translation, i must use perl or shell script?

There are differences in output, for example:

perl output:
```
Listing processes running inside tomb '::1 tombname::'...
```
shell output:
```
Listing processes running inside tomb '::1 tombname::'..." "$1
```

There are worst problems than that.

jaromil:
Of the two choices above, the perl output looks like the correct one. I am not sure what other problems are there.

However, we used weblate for tomb, I just created a BR_pt one there
https://hosted.weblate.org/projects/tomb/tomb/pt_BR/

jaromil:
[tomb-tomb-pt_BR.zip](https://github.com/dyne/Tomb/files/11884097/tomb-tomb-pt_BR.zip)

here is the .PO file in case you don't like to use weblate

PR #483, opened 2023-06-26
Tomb accepts also plocate for search/index (not only mlocate)

nerun:
### Related issue: #464

# The problem
Tomb is an old tool and uses `mlocate` for searching/indexing. But modern systems have replaced `mlocate` with `plocate`. On these systems (such as Linux Mint 21.1 and Ubuntu 22.04) the `tomb index` or `tomb search` commands are broken, user receives message:
```
Index command needs 'mlocate' to be installed.
```
To fix this issue i have changed lines with:
```
[[ "$updatedbver" =~ "mlocate" ]] || {
    _failure "Index command needs 'mlocate' to be installed." }
```
To:
```
[[ "$updatedbver" =~ "locate" ]] || {
    _failure "Index command needs 'mlocate/plocate' to be installed." }
```
Now both systems (with mlocate or plocate) works fine.

# Difference between mlocate and plocate
[plocate](https://plocate.sesse.net/) is backwards-compatible with mlocate, and is much faster and more efficient than mlocate.

In fact you don’t have a choice between mlocate and plocate in Ubuntu 22.04: if you install the former, you’ll end up with the latter anyway, because [mlocate is a transitional package](https://packages.ubuntu.com/jammy/mlocate) which pulls in [plocate](https://packages.ubuntu.com/jammy/plocate).

Source: [Difference between mlocate and plocate](https://unix.stackexchange.com/questions/727862/difference-between-mlocate-and-plocate)

### DIFF
--- a/tomb
+++ b/tomb
@@ -2813,13 +2813,13 @@ BEGIN { }
 # $1 is optional, to specify a tomb
 index_tombs() {
 	{ command -v updatedb 1>/dev/null 2>/dev/null } || {
-		_failure "Cannot index tombs on this system: updatedb (mlocate) not installed." }
+		_failure "Cannot index tombs on this system: updatedb (mlocate/plocate) not installed." }
 
 	updatedbver=`updatedb --version | grep '^updatedb'`
 	[[ "$updatedbver" =~ "GNU findutils" ]] && {
 		_warning "Cannot use GNU findutils for index/search commands." }
-	[[ "$updatedbver" =~ "mlocate" ]] || {
-		_failure "Index command needs 'mlocate' to be installed." }
+	[[ "$updatedbver" =~ "locate" ]] || {
+		_failure "Index command needs 'mlocate/plocate' to be installed." }
 
 	_verbose "$updatedbver"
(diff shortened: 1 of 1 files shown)


jaromil:
Hi nerun!

I see this PR just changes error messages, which is OK if we can be sure to rely on distros to keep aliasing executables as locate/updatedb.

One request before merging, which should be easy to do: can you please also update the .po translations with this detail? so we keep them aligned.

 thanks!

nerun:
@jaromil , about PO files, no problem! But could you help me with issue #482 ? It's related to translations and PO files.

This PR specifically doesn't just fix error messages. Index/search commands don't work in Linux Mint (and probably Ubuntu 22.04). Precisely because of the mlocate problem. These commands specifically look for mlocate, when they don't find it they don't work and return the error message. What this PR does is tell the index/search commands to look for mlocate `OR` plocate. And then everything works on Linux Mint/Ubuntu (current versions).

nerun:
I suggest merge this PR `before` merge PR #483.

PR #491, opened 2023-09-30
Fix operators to compare numerically, not lexicographically

### LINKED ISSUE #489: File tomb, function lock_tomb_with_key(): Wrong comparison operator used to check a tomb's size
The operator `<` compares two strings lexicographically resulting in that a 100MB tomb is considered smaller than 47MB or 18MB.

Please consider this patch from Debian to fix the issue:
https://salsa.debian.org/pkg-security-team/tomb/-/blob/debian/master/debian/patches/comparison-operator.patch

Narrat:
The operator < compares two strings lexicographically resulting in that a 100MB tomb is considered smaller than 47MB or 18MB.

Closes #489

__________
Took the liberty and created the PR which incorporates the fix for #489

### DIFF
--- a/tomb
+++ b/tomb
@@ -2162,13 +2162,13 @@ lock_tomb_with_key() {
 			ext3|ext4) ;;
 			ext3maxinodes|ext4maxinodes) ;;
 			btrfs)
-				if [[ $tombsize < 49283072 ]]; then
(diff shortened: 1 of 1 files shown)


jaromil:
Many thanks for this and the other PR, please take all the liberties 😅

Issue #433, opened 2022-02-13
Space characters in bind-hooks


zhogano:
How can I implement bind-hooks if there is a space character in the path relative to $HOME? I enclosed the path names with "" but it did not work. Then I tried using symlink. Tomb opened without problem, but I was unable to close it with `tomb close` and `tomb slam`. I am using Tomb 2.9.0. I can not remove the space character from the folder name, because it was created by a software and it won't work if I change it.

jaromil:
Hi! can you please confirm this issue is solved using latest version from git?

zhogano:
Hi @jaromil. It did not work for me. I downloaded the latest version from git, installed it using `sudo make install`. Made sure the file "tomb" changed in `/usr/local/bin/`. 
Let's say this is the path  `/home/user/.config/Example Folder` which I want to bind-hook to `media/ExampleFolder` in tomb.

1)  First I tried without symlink. But it could not bind-hook:
bind-hooks:
`media/ExampleDirectory ".config/Example Directory"`

Output:
`tomb  .  Commanded to open tomb .doc/secret.tomb
tomb  .  Valid tomb file found: .doc/secret.tomb
tomb  .  Key is an image, it might be valid.
tomb  .  Mountpoint not specified, using default: /run/media/user/secret
tomb (*) Opening secret on /run/media/user/secret
tomb  .  This tomb is a valid LUKS encrypted device.
tomb  .  Cipher is "twofish" mode "xts-plain64" hash "sha512"
tomb  .  A password is required to use key .....
tomb [W] Attempting key recovery.
tomb  .  Password OK.
DM-UUID for device tomb.secret.d52da4cfad13566f9c357890434340a54752285bd92d14589740aedc17671561.loop14 was truncated.
tomb (*) Success unlocking tomb secret
tomb  .  Filesystem detected: ext4
tomb  .  Checking filesystem via /dev/loop14
fsck from util-linux 2.37.4
secret: clean, 24238/3399680 files, 10277943/13567488 blocks
tomb (*) Success opening secret.tomb on /run/media/user/secret
tomb  .  Last visit by user(1000) from /dev/pts/1 on user-pc
tomb  .  on date Mon 07 Mar 2022 10:36:37 AM +03
tomb [W] bind-hook target not existent, skipping /home/user/".config/Example`

2) Then I tried using symlink. This way the bind-hook worked, but then it was impossible to close the tomb either by `tomb close` or `tomb slam`.

I created symlink with
[...]

PR #500, opened 2023-12-15
Fix spaces in bind-hooks paths

Llewellynvdm:
https://github.com/dyne/Tomb/blob/a0b633fb774626e6f6e9767b70247bf5490dd0e2/tomb#L2593-L2604

The option to target folders/files that have spaces in the name would be ideal, so I added a little work around.

### DIFF
--- a/tomb
+++ b/tomb
@@ -2594,8 +2594,11 @@ exec_safe_bind_hooks() {
 	# each line, using zsh word separator array subscript
 	_bindhooks="${mapfile[${mnt}/bind-hooks]}"
 	for h in ${(f)_bindhooks}; do
+		h=${h//\\ /__ESC_SPACE__}
 		s="${h[(w)1]}"
 		d="${h[(w)2]}"
+		s=${s//__ESC_SPACE__/ }
+		d=${d//__ESC_SPACE__/ }
 		[[ -z $s ]] && { _warning "bind-hooks file is broken"; return 1 }
 		[[ -z $d ]] && { _warning "bind-hooks file is broken"; return 1 }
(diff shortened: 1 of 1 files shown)


Llewellynvdm:
This should also resolve https://github.com/dyne/Tomb/issues/433

Llewellynvdm:
@jaromil is there a particular reason why this PR is in limbo?

jaromil:
Hi! Many thanks for your PR! It not in limbo, just some delay due to new year's holiday and the fact this project is ran by volunteers.

I am just wondering if the renaming to `__ESC_SPACE__` is appropriate, or shall we just use a `_` underscore for the generated targets used by bind mount.

Llewellynvdm:
The use of `__ESC_SPACE__` as a placeholder in the script serves a specific purpose: it is intended to be a unique sequence that is unlikely to occur in actual file or directory names. This uniqueness is essential because the placeholder temporarily represents spaces in paths during processing to avoid issues with word splitting.

If we were to use a single underscore `_`, there is a real possibility that actual path names could contain underscores. These would then be erroneously converted to spaces when the placeholder is reverted, leading to incorrect path names. By using `__ESC_SPACE__`, we mitigate this risk because it is designed to stand out and not conflict with typical naming conventions.

Therefore, while the placeholder may seem verbose or unconventional, its distinctive pattern is a safeguard to ensure that only the intended spaces are replaced and then accurately restored, preserving the integrity of the file paths during the bind mount process.

jaromil:
Excellent explanation, thanks for all your attention to details and welcome among the Tomb contributors!

PR #501, opened 2023-12-16
Use cat instead of _cat

0rtz:
_cat is not defined/autoloaded anywhere, so use plain unix cat utility to avoid getting "command not found" error


jaromil:
Hi! and good catch! _cat is a leftover from previous use of zsh builtin functions, now dismissed.

PR #506, opened 2024-01-10
Fixed tomb listing failure message: localization friendly.

nerun:
When listing tombs with `tomb list`, if there are no tombs open, this failure message is shown:

Line 2668:
```shell
_failure "I can't see any ::1 status:: tomb, may they all rest in peace." ${1:-open} }
```

The problem is that the `::1 status::` is always `open`. So the message shown will be always:

```
I can't see any open tomb, may they all rest in peace.
```
But the word `open` cannot be translated because it is a variable. In other languages, it sounds bad.

The phrase `I can't see any open tomb, may they all rest in peace.` is used twice along tomb script, in lines 2832 (`index_tombs`) and 2946 (`search_tombs`). So we can safely adopt this phrase also for `list_tombs` in line 2668.

### DIFF
--- a/extras/translations/tomb.pot
+++ b/extras/translations/tomb.pot
@@ -1,6 +1,6 @@
 
 # Tomb - The Crypto Undertaker.
-# Copyright (C) 2007-2014 Dyne.org Foundation
+# Copyright (C) 2007-2024 Dyne.org Foundation
 # Denis Roio <jaromil@dyne.org>, 2013.
 #
 #, fuzzy
@@ -1123,7 +1123,7 @@ msgid "Tomb ::1 version:: - a strong and gentle undertaker for your secrets"
 msgstr ""
 
 #: tomb:Main routine:main:2806
-msgid " Copyright (C) 2007-2017 Dyne.org Foundation, License GNU GPL v3+"
+msgid " Copyright (C) 2007-2024 Dyne.org Foundation, License GNU GPL v3+"
 msgstr ""
 
 #: tomb:Main routine:main:2807


Sakrecoer:
hi and well met! Is this a transalation issue? If so: wich language are you reffering to?

nerun:
Yes and no. Brazilian portuguese (pt_BR).

In pt-br, `open` is translated as `aberto`. But variables cannot be translated. So, in any other language not just pt-br, `open` will not be translated. That's the purpose of this PR. In english will not make difference.

Sakrecoer:
Wait, are you saying the translation of the variable `open` stays "open", even if it's another language? (öppen, ouvert, abierto, offen..)

My daily language varies, but my robots speak strictly English. Sorry if i misunderstood.

Sakrecoer:
I am so sorry. I shouldn't be practicing my fluorescent balck belt in derp-fu here. Yet, i remain triggered by language. (In my Aikido; the notification i got put me under the impression this was an _issue_, about language. But isee now it is a PR)

Neat unearthing, nevertheless! Let's hope it passes!

nerun:
No. Please read the commit to understand it better.

`::1 status::` is the variable used in the failure message, but it is always "open". But `::1 status::` is used in "po" files. What i suggest it tobuse "open" directly in the message, not a variable.

jaromil:
shouldn't we change the .pot file then and update some translations? perhaps this can be done inside this PR

nerun:
Yes, you're right! I will update the POT. This also raises other questions though.

As I explained in  [extras / translations / README](https://github.com/dyne/Tomb/tree/master/extras/translations) ALL `.po` files are outdated (except pt_BR-2.10.po). They are for Tomb 2.3 not 2.10+. I created that `tomb-2.10.pot` for this reason in a very old PR already merged if you remember, but the old `tomb.pot` (for v2.3) was not deleted. I think we should delete that one too, and rename the `tomb-2.10.pot` to just `tomb.pot`. Will be a good idea to rename ALL the translation `po` files to `<language>_outdated_v2.3.po`.

What do you think?

nerun:
Moved to PR #512

PR #507, opened 2024-01-10
manpage: typo fix, column formatting, .EX/.EE macros indentation etc

nerun:
Under `bind-hooks`, in the line:
```
the tomb. The files and directories will be be made directly
```
Word `be` appears twice.

And in PASSWORD INPUT, added linke breaks to fit 80 columns. It was a "very long line". Just keeping the document standard.

Better formatting of .EX/.EE macros adding indentation.

Replaced `mlocate` with `mlocate/plocate`. Since commit 59d7331 added plocate as a possibility.


melroy89 (approved):

jaromil:
thanks

PR #525, opened 2024-07-22
add editorconfig and lint tomb script

jaromil:
Changes to normalize the format of tomb sourcecode according to a shipped editorconfig definition

adopted tab style with 8 spaces


Narrat:
Why size 8 and not 4 for the real tabs? It stretches greatly... 
Mostly curious, as I can locally adjust to size 4 :D

PR #531, opened 2024-08-01
Optim/ci

Narrat:
While adding argon2 to workflow file I noticed some details

1) folder portable got moved
2) the tests for pbkdf2 should maybe be run before using them with the tomb tests


Narrat:
Or the workflow for the portable should be dropped entirely

jaromil:
We can archive the workflow of portable rewrite in its extras folder. I lost interest in completing it, seems noone is really interested. I'll simply make sure that the tomb script works on busybox.

Narrat:
Alright. Will move the file

PR #534, opened 2024-08-06
Man: merge section on Password Input

Narrat:
Somehow this section existed two times in the man page with similar information.
_____________________________
Dunno how accurate the comment in regards to `gpg-agent` is (especially the last sentence).
But I adjusted it somewhat in regards to Wayland and the EOL of GTK2 which makes `pinentry-gtk2` not the first choice anymore. Some distributions are actively working on removing GTK2 entirely.

### DIFF
--- a/doc/tomb.1
+++ b/doc/tomb.1
@@ -424,12 +424,23 @@ To avoid that tomb execution is logged by \fIsyslog\fR also add:
 
 Password input is handled by the pinentry program: it can be text
 based or graphical and is usually configured with a symlink. When
-using Tomb in X11 it is better to use a graphical pinentry-gtk2 or
-pinentry-qt because it helps preventing keylogging by other X
-clients. When using it from a remote ssh connection it might be
-necessary to force use of pinentry-curses for instance by unsetting
-the DISPLAY environment var.
+using Tomb in a graphical environment (X11 or Wayland) it is better
+to use either pinentry-gtk2 (deprecated), pinentry-gnome or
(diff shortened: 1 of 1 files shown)


doc/tomb.1
```diff
@@ -424,12 +424,23 @@ To avoid that tomb execution is logged by \fIsyslog\fR also add:
...
-using Tomb in X11 it is better to use a graphical pinentry-gtk2 or
-pinentry-qt because it helps preventing keylogging by other X
-clients. When using it from a remote ssh connection it might be
-necessary to force use of pinentry-curses for instance by unsetting
-the DISPLAY environment var.
+using Tomb in a graphical environment (X11 or Wayland) it is better
+to use either pinentry-gtk2 (deprecated), pinentry-gnome or
+pinentry-qt because it helps preventing keylogging by other clients.
+When using it from a remote ssh connection it might be necessary to
+force use of pinentry-curses for instance by unsetting the DISPLAY
```
jaromil, line 431: or pinentry-tty, actually even better in these situations.

jaromil:
I think we can exclude gpg-entry to become mandatory and also the manual starting, belongs to its own manpage.

Thanks BTW well spotted.

doc/tomb.1
Narrat: Okay, will change that

Narrat:
Okay. Will remove those parts then

jaromil:
cheers

PR #538, opened 2024-08-31
enable doas tests and add argon2 to CI

jaromil:
fix some test units and minor cleanups


Narrat:
Only noticed now. The sphinx removal needs a small follow-up.
The options are still mentioned in the subcommands
https://github.com/dyne/tomb/blob/master/tomb#L3128 (forge, setkey, passwd, bury and exhume)

Issue #542, opened 2024-09-18
Pinentry fallback mechanism is poorly constructed


ultimatespirit:
Checked the issue history and saw several pinentry issues that basically went over this exact same issue and never really got handled in any longterm manner.

# Issue

As noted [here on line 491](https://github.com/dyne/tomb/blob/ff692999de9fdec625ffab60c57f3cc6cf1ccd16/tomb#L491) tomb rolls its own fallback mechanism for `/usr/bin/pinentry`, and does so poorly.

## Current algorithm:
- Check if a `DISPLAY` is set
- If so, check if a specific command happens to exist, and if it does **use it unconditionally**
- Else, try more commands
- Else, use `pinentry-curses` else `pinentry-tty` (one of those virtually guaranteed to succeed if they exist)
- Else, fail

## Algorithm used by `/usr/bin/pinentry` wrapping script
Note, this is from my Arch Linux installation with pinentry 1.3.1-5.
```bash
# Guess preferred backend based on environment.
backends=(curses tty)
if [[ -n "$DISPLAY" || -n "$WAYLAND_DISPLAY" ]]; then
	case "$XDG_CURRENT_DESKTOP" in
	KDE|LXQT|LXQt)
		backends=(qt qt5 gnome3 gtk curses tty)
		;;
	*)
		backends=(gnome3 gtk qt qt5 curses tty)
		;;
	esac
fi

for backend in "${backends[@]}"
do
	lddout=$(ldd "/usr/bin/pinentry-$backend" 2>/dev/null) || continue
#       ^~~~ most important part here, emphasis mine
	[[ "$lddout" == *'not found'* ]] && continue
	exec "/usr/bin/pinentry-$backend" "$@"
done
```
Which importantly _checks that the found executable would even run_ before deciding on it.

# Proposed solution
Several fold solution:
1) Just adopt that wrapper script's implementation. Checking an array of potential executables is clearly more maintainable than the current hand-unrolled implementation, and `ldd` is guaranteed to be on any functional system, or at least any sane one. On the off chance it's a musl-only system _and_ they didn't make a `ldd` symlink, consider having support for `ld-musl-$ARCH.so`.
2) Provide a user-facing option to just specify the pinentry they want to use at command invocation time
3) Also, add `pinentry-gtk` to the list of potential applications, with gtk2 gone Arch Linux at least has removed `pinentry-gtk2` and made `pinentry-gtk3` -> `pinentry-gtk`

# The implementation
```bash
[...]

PR #543, opened 2024-09-18
Rewrites `pinentry` find code to be more resilient

ultimatespirit:
This change rewrites the custom `pinentry` search code to instead be a modified form of the standard `/usr/bin/pinetry` fallback script. The prior behaviour could not handle cases where a `pinentry` executable existed but was not actually usable. Now it checks using `ldd` for if an executable is functional or not. Additionally rewritten to be more clear and easier to extend with newer `pinentry` backends.

`make test` was ran, and it passed the main usability tests for password entry. As my testing system uses swap it failed during the KDF test from an error from swap being enabled; I did not attempt further tests with swap disabled.

This partially fixes Issue #542.

Note: This adds a new dependency on `ldd`. For systems without `ldd` installed (such as `musl` systems) they either need to create a compatibility `ldd` symlink / script, or we need to check for `ld-musl-$ARCH.so` and use it ourselves if `ldd` does not exist.

### DIFF
--- a/tomb
+++ b/tomb
@@ -487,12 +487,17 @@ ask_password() {
 	local gtkrc
 	local theme
 	local pass_asked
+	local backends
 
 	# Distributions have broken wrappers for pinentry: they do
 	# implement fallback, but they disrupt the output somehow.	We are
-	# better off relying on less intermediaries, so we implement our
-	# own fallback mechanisms. Pinentry supported: curses, gtk-2, qt4, qt5
-	# and x11.
+	# better off relying on less intermediaries, so we copy in the
+	# fallback logic directly here.
+	# Pinentry supported: curses, tty, qt{,4,5}, gtk{,-2}, gnome3, and x11.
+
+	# TODO: Implement a user option to specify a pinentry program and
+	# wrap pinentry with a `_pinentry()` function that implements the
+	# search logic if the user option is not specified: Issue #542
 
 	# make sure LANG is set, default to C
 	LANG=${LANG:-C}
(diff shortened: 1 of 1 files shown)


jaromil (changes requested):
It's a neat improvement, good to ldd check. But we need to build the musl compatible alternative in it to avoid breaking people's workflow.

Narrat:
It is documented in the official [FAQ](https://wiki.musl-libc.org/faq#Q:-Where-is-%3Ccode%3Eldd%3C/code%3E?) so users should be aware of such specialities and take care of it themselves? Magically doing stuff behind the scenes like tampering with `$PATH` and files therein is kinda not nice.

ultimatespirit:
Sorry, been busy with work and haven't had time to address this yet. Going to leave a quick proof sketch, and a follow up.

First some administrative points:
- I do not have a musl system set up, so I cannot test this directly.
- I do not have a non-x86_64 system set up, nor is it running musl, so I cannot test _that_ directly either.
- When I tried doing some further research into the practicalities of detecting musl usage, beyond just the FAQ about ldd in musl we all know, I saw repeated mentions of systems with musl having a `ldd` in their path anyway, presumably from distributions running musl knowing to implement that sanity work-around. So this may be superfluous in the expected cases anyway.
- I'd appreciate testing by people who do have musl, or non-x86_64 architectures and musl, either of the proof sketch below or the final patch whenever I get to make it.
- P.S. I keep mentioning "make a CLI argument to override this" for the implementation, in part the reason I haven't just _done_ that is due to making this PR to fix my immediate breakage on archlinux, and not really wanting to read through the semi-complicated looking CLI parsing logic I saw in the `tomb` script in the bottom. The actual concept isn't too hard, just get a `_pinentry_cmd` variable defined by a cli arg, check if it exists, and if so skip all the checking logic. If not, do the logic implemented here for determining what `pinentry` to use.

Basic idea:
```bash
# `_ldd` must ALWAYS be used in a subshell, it potential runs `exec`.
local _ldd
local _potential_musl="/lib/ld-musl-$(uname -m).so.1"
if _is_found ldd; then
    # Nothing to do
    _ldd="ldd"
elif [ -f "$_potential_musl" ]; then
    # Musl will act like `ldd` if called as `ldd`. This requires an exec hack sadly.
    _ldd="exec -a ldd $_potential_musl"
else
    # Give up and revert to old behaviour of just blindly using the first thing we find
    # TODO: Make a CLI arg to specify pinentry executable and mention it in this warning
    _warning "System has no 'ldd' command nor does it appear to be running musl, falling back to imprecise pinentry detection mode"
    _ldd="true"
fi
unset _potential_musl
...
lddout=$($_ldd "$pinentry" 2>/dev/null) || continue
[[ "$lddout" == *'not found'* ]] && continue
```
The `uname` and `true` commands are specified by POSIX, if concerned could use `:` instead of `true` but I feel that's less clear to casual readers. Note that we don't actually need to use `true` here, since the detection is string reading based anyway, so long as the command used doesn't output the line "not found" it's a valid fall through case.

jaromil (approved):

jaromil:
The `_pinentry_cmd` variable defined by a cli arg to override detection is a nice idea. Ideally any automatic detection should have a way to be overridden.

jaromil:
Time to merge this! very well done

PR #557, opened 2025-01-27
add Github action for syncing translation files

Narrat:
POC of using Github Actions to automatically sync translations files with the base script

Tested on my personal fork (poor main branch :D ):
[Log of successful run](https://github.com/Narrat/Tomb/actions/runs/12996864257)
[Generated PR](https://github.com/Narrat/Tomb/pull/2)

Just as a draft for the general idea. The script is very barebones and my experience with GHA is minimal.
First and foremost done to see what you think of this :)

### DIFF
--- a/.github/workflows/sync_pot.yaml
+++ b/.github/workflows/sync_pot.yaml
@@ -1,0 +1,29 @@
+name: Translations - sync translation files with tomb script
+
+on:
+  push:
+    branches: master
+    paths:
+      - 'tomb'
+
+jobs:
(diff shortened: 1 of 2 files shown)


Narrat:
After pondering on this for some time I think for initial adoption it should be this barebone.
It is for an extra and if it fails it should fail early. And especially since Weblate still is the main place for translations (although currently disabled). But I hope with adding something like this, it will make it easier to keep both repos in think without weblate breaking down.

Issue #34, opened 2011-08-22
Close tombs on hibernation / suspend


boyska:
Hibernating while a tomb is open is bad: it means that your key is written to disk.
We should _close_ every tomb before hibernating.
This can be done using /etc/pm/sleep.d/99-tomb script which does this.

Also, if closing fails (because of pid locking it) we should choose between slamming and preventing hibernation.
Slamming is easier, but could be unintuitive for the user to "hibernate" and not finding his applications open on reboot.
Preventing hibernation seems to be possible, according to this post: http://askubuntu.com/questions/28328/how-can-i-disable-hibernate-completely-in-kubuntu

CameronNemo:
In addition to a script in `/etc/pm/sleep.d/`, you will want another script in `/lib/systemd/sleep/` for systemd-logind setups.

jaromil:
Any tip for examples or templates or existing scripts in distributions? I'd go for something already tested rather than reinvent the wheel here.

Narrat:
An example for a systemd unit got posted in #310

Narrat:
The next issue I want to look at. Which either is a two parts or just one.
First I want to check how feasible it would be to add two new subcommands (like `sleep/slumber` and `revive`) which make use of `luksSuspend` and `luksResume`. Using those outside of `tomb` is not really feasible, as `luksResume` requires the password again. Which is not as simple as a normal password in the case with tomb.
And it needs some cover mount operation, else the cached top level values from the mount would still be visible. Although interacting with those won't be possible at all.
Thinks that needs a look:
- [ ] Is a temporary cover mount on top even possible
- [ ] can suspend and resume safely be done if the mount is still there (`luksSuspend` does work; `luksClose` does not work (still in use), ; `umount` on suspended doesn't work; `luksResume` not really tested, but it wanted the password :D )
- [ ] how do programs with open files respond to the device being suspended (aka would it be even safer/better to use `slam`)

Secondly a section in extras which adds unit files for the various init systems.
Depending on the result from the first part this is additionally or the remaining solution to "support" automatic close on hibernation.

Narrat:
POC PR inbound.
Generally:
* luksSuspend & luksResume work great
* tmpfs mount on top hides cached values
* in general programs trying to work on the suspended device will hang until it is resumed

[...]

PR #558, opened 2025-01-29
add support for luksSuspend/luksResume

Narrat:
May come in handy for hibernation/suspending the OS without completly closing the tomb.

___________________
Just a POC for #34 without a full close operation, but still avoiding writing the secret to places where it shouldn't be.
Not fully fleshed out, as bind mounts are ignored, operations on a suspened device are allowed.
But it is imo enough to play around with and decide if it is worth or close (or slam) is enough.

### DIFF
--- a/tomb
+++ b/tomb
@@ -2935,6 +2935,160 @@ resize_tomb() {
 
 # }}}
 
+# {{{ Suspend & Resume
+
+# ToDo
+# * what should be done with bind-hooks?
+# * what about access of files when suspending (aka kill everything needed?)
+# * tomb operations on a suspended mount
+#   -> extra check for all operations, if device is suspended
+#     -> # cryptsetup status $mapper
+#     -> (...)
+#     -> mode: read/write (suspended)
+#     -> YES, for close/slam
+#     -> NO for list
+#     -> ? for passwd
+suspend_tomb() {
+	local tombs how_many_tombs
+	local pathmap mapper tombname tombmount loopdev
+	local ans pidk pname
+
+	if [ "$1" = "all" ]; then
+		mounted_tombs=(`list_tomb_mounts`)
+	else
+		mounted_tombs=(`list_tomb_mounts $1`)
+	fi
+
+	[[ ${#mounted_tombs} == 0 ]] && {
+		_failure "There is no open tomb to be suspended." }
+
+	[[ ${#mounted_tombs} -gt 1 && -z "$1" ]] && {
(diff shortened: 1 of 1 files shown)


tomb
```diff
@@ -2935,6 +2935,160 @@ resize_tomb() {
...
+	tombname=$1
+	mounted_tombs=(`list_tomb_mounts $tombname`)
+	for t in ${mounted_tombs}; do
+		mapper=`basename ${t[(ws:;:)1]}`
+		tombmount="${t[(ws:;:)2]}"
+		_verbose "Mapper: ::1 mapper::" $mapper
+	done
+
+	_message "Commanded to revive tomb ::1 tomb name::" $tombname
+	# Currently missing check if $tombname is valid or could be found
```
Llewellynvdm, line 3052: Can this work?
```
if [[ ${#mounted_tombs[@]} -eq 0 ]]; then
	_failure "No mounted tombs found for: $tombname"
fi
```

tomb
```diff
@@ -2935,6 +2935,160 @@ resize_tomb() {
...
+		_failure "Failure resuming the encrypted mapper." }
+
+	_success "Success reviving tomb ::1 tomb name::" $tombname
+
+	# process bind-hooks (mount -o bind of directories)
+	# and exec-hooks (execute on open)
+	#option_is_set -n || {
+	#	exec_safe_bind_hooks "${tombmount}"
+	#	exec_safe_func_hooks open "${tombmount}"
+	#}
```
Llewellynvdm, line 3085: Not sure if this is what we are going for
```
# Process bind-hooks and exec-hooks safely
if ! option_is_set -n; then
	if [[ -d "$tombmount" ]]; then
		exec_safe_bind_hooks "$tombmount" || _failure "Failed to execute bind-hooks."
		exec_safe_func_hooks open "$tombmount" || _failure "Failed to execute function hooks."
	else
		_failure "Invalid tomb mount directory: $tombmount"
	fi
fi
```

tomb
Narrat: As a general check, yes. But it would still need to check on available/open mounts and their names in detail.

tomb
Narrat: The right direction.
I didn't look into implementing this, as the general question still needs an answer.
Does is offer some kind of benefit? The script would contain three ways to "close" a tomb. `close`, `slam` and then `suspend/resume`. With `suspend/resume` kinda in between `close` and `slam` and I'm still not sure it is a benefical addition.

PR #550, opened 2024-11-11
Translations update from Hosted Weblate

weblate:
Translations update from [Hosted Weblate](https://hosted.weblate.org) for [Tomb/tomb](https://hosted.weblate.org/projects/tomb/tomb/).

Current translation status:

Weblate translation status

PR #561, opened 2025-03-11
Weblate merge conflict

Narrat:
Replaces #550
Manual merge with resolved conflicts.
Should resolve the stuck situation with the diverged repos.


Narrat:
If that resolves the situation, #557 should likely be of short term priority to avoid desync of translations.

PR #564, opened 2025-04-14
_is_found(): make use of the function

Narrat:
instead of the repeated command -v calls.
__________
I noticed this function, but did not check why it wasn't used.
So either it should be used or dropped :)


jaromil:
Yes, I created it and then forgot to use it everywhere. The reason to have it is that we may need to change from `command -v` to something else in some cases, and then there is no need to change the code everywhere. So far however it seems every platform supports `command -v`. I am in favor of merging this in any case, because it makes the code more readable.

Narrat:
I see :) 
One remark on the merge. This partially reverted the change from #563. I assume this wasn't the intention? If so a small commit needs to follow :) Would get around to that later the day

Issue #575, opened 2025-07-24
v2.12 rejects password for all key files


stevearc:
After the most recent update I have found the following issue:

1. `./tomb open archive.tomb -k archive.tomb.key`
2. Prompted in terminal for sudo password (works fine)
3. Opens GUI popup to prompt for key password
4. Rejects password, asks a second time
5. Rejects password, asks a third time
6. Fails with error message `tomb [E] No valid password supplied.`

I ran a git bisect and traced the issue back to #543

If I do a straight revert on master `git revert 9f79e8b` it works again (accepts my password from the gui on the first attempt).

I'm using KDE on arch

Narrat:
Thank you for reporting and bisecting the issue.
This reminds me of a experience I had with WSL2 Arch and openSUSE Tumbleweed. In those cases the TTY/ncurses pinentry appears but the same happens: Rejects password for the key.
And it worked before. Sadly I did not investigate it further, as it was elsewhere working. I lazily assumed some changes on WSL side... Guess what came back biting :D 

Can you provide the ouput for the following env vars?
```
DISPLAY
WAYLAND_DISPLAY
XDG_CURRENT_DESKTOP
```
Would it possible to provide a debug output for a test tomb? The CLI option would be `-D`

Finadil:
Same issue trying to open a tomb. Arch Linux KDE Plasma (Wayland). Even trying `tomb forge asdf.key` it just keeps asking for a password.
For debugging I modified the end of ask_password() to add verbose output:
```
	_verbose "password: $password"
	print "$password"
	return 0
```
and in gen_key:
```
	tombpass=`ask_password "Type the new password to secure your key"`
	_verbose "tombpass: $tombpass"
```
The output was:
```
❯ tomb forge asdf.key -D
tomb [D] Identified caller: finadil (1000:1000)
tomb [D] Updating HOME to match user's: /home/finadil (was /root)
tomb [D] Tomb command: forge asdf.key
tomb [D] Caller: uid[1000], gid[1000], tty[/dev/pts/4].
tomb [D] Temporary directory: /tmp
tomb  .  Commanded to forge key asdf.key with cipher algorithm AES256
tomb [W] This operation takes time. Keep using this computer on other tasks.
tomb [W] Once done you will be asked to choose a password for your tomb.
tomb [W] To make it faster you can move the mouse around.
tomb [W] If you are on a server, you can use an Entropy Generation Daemon.
tomb [D] Data dump using dd from /dev/urandom
[...]

PR #576, opened 2025-07-28
Address unexpected output to stdout in ask_password()

Narrat:
Contains the fix for #575 plus some cleanup to `ask_password()`.
What every changeset does is best retrieved from the respective commit messages.
As this can affect a lot of people I suggest a new release after that issue got addressed. For example every distribution with a qt6 based DE will run into this, as `pinentry-qt5` may not be available.

### DIFF
--- a/tomb
+++ b/tomb
@@ -522,7 +522,7 @@ ask_password() {
 				;;
 		esac
 	fi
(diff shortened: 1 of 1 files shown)


jaromil:
many thanks @Narrat  owe you a beer or more

PR #580, opened 2025-08-20
Add option --pinentry

Narrat:
This allows to specify a specific pinentry program which is tried first while finding a working pinentry program.
For input is only the part needed after "pinentry-". This isn't a general option to specify something completly different as password input (yet).

The update to the manpage includes the new option and an update to the pinentry section. With the way how pinentry is nowadays detected, the manual force of a specific pinentry via symlink shouldn't be encouraged. And if a specific one should be used, one can use now the new option.
____________
Doesn't really help with the big ToDos in `ask_password()`, but allows in theory the usage of pinentries like `wayprompt` or `pinentry-bemenu`.
Did take a look into adding the wrapper.. but this is not a small change. Especially with all the possible corner cases.
And with the ToDo in regards to use `read -s -r` as a fallback... this needs some more thinking.

### DIFF
--- a/doc/tomb.1
+++ b/doc/tomb.1
@@ -329,6 +329,12 @@ Select a different tool than sudo for privilege escalation.
 Alternatives supported so far are: pkexec, doas, sup, sud. For any
 alternative to work the executable must be included in the current
 PATH.
+.B
+.IP "--pinentry \fI<toolkit>\fR"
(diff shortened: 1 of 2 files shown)


jaromil:
Oh this is very good for now, should have been there since the beginning: always provide a way to select manually the options.

PR #584, opened 2025-09-04
Rework logic in `_sudo()` and add support for  `sudo-rs`

Narrat:
The `sudo` wrapper `_sudo` has different behaviour depending if `--sudo` is  used or not. It is more chatty with `-D`, `--sudo sudo` doesn't work and code duplications which could theoretically be avoided.
Therefore an attempt to make the function more general and instead of the two branches just provide one. The logic itself is inspired by the structure from the `_pinentry` wrapper.

And add support for [`sudo-rs`](https://github.com/trifectatechfoundation/sudo-rs)

Not sure adding the verbosity in general... It is rather chatty, but also a reminder for all those places were privilege escalation is needed :D

### DIFF
--- a/doc/tomb.1
+++ b/doc/tomb.1
@@ -325,8 +325,8 @@ takes massively longer on regular systems.
 Default is 1 thread (based on the \fIargon2\fR default).
 .B
 .IP "--sudo \fI<executable>\fR"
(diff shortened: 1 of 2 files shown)


jaromil:
Excellent! much more readable now too.

PR #586, opened 2025-10-03
Change password input for gpg from STDIN to dedicated fd

### LINKED ISSUE #304: My buried data has been sucked inside the liquid core of the planet
Hi all,

I made the bad mistake of putting a backslash in my password. I debugged a bit the problem and found two separate unexpected behavior (at least, I wasn't expecting them ;-) ). Please let me know if my interpretations are incorrect. Basically, backslashes are not escaped. This leads to two different consequences.

1. A backslash inside the password, not being escaped, is used to form ascii control (non-printable) characters; for example \t enters in the password as tab (0x9), \f as form feed (0xC), and so on.

2. A backslash at end of the password has more catastrophic effect (as far as I understood), neutralizing the '\n' in gpgpass="${tombpass}\n$TOMBSECRET", so causing gpgpass to be a concatenation of the password with the generated key, which if I understood well means that the gpg passphrase for the key is key itself, and i don't come up with any way to decrypt it (didn't make calculations but I think that brute-forcing it could imply the usual life-time of the universe).

Point (2) is what happened to me, and sadly, since this tomb was already in use with another password, I lost some important stuff. So, let me know if there are any mistakes in my reasoning.

The issue should be easily reproducible, but let me know if you need some details of my system.

That said, tomb is an utterly astonishing tool, and thank you very for your good work.

### LINKED ISSUE #581: Drop workaround for `gnupg 1.4.x`
The 1.4 release of gnupg is long time deprecated and according to upstream shouldn't be used anymore.
Various distros still have the package for compatibility reasons, but even Debian switched the name of the binary to [`gpg1`](https://packages.debian.org/sid/amd64/gnupg1/filelist). Even the case with [Bullseye](https://packages.debian.org/bullseye/gnupg1).
So in theory the majority should already use `gpg2` as default.
Is there any distro around, which still ships `gpg1` as `gpg`? (Hopefully not)

Reason being that it would allow to drop the gpg1 specific [workaround](https://github.com/dyne/tomb/blob/master/tomb#L1145).
With that gone the `gpg` call could be restructured without worries about breaking something.

What could be changed? There is still #304 around. And for the `gpg` call password and key are still combined via `\n` and piped into it.
There are in theory three options to retrieve a password with two being considered questionable
```
   --passphrase-fd n
          Read the passphrase from file descriptor n. Only the first line will
          be  read  from file descriptor n. If you use 0 for n, the passphrase
          will be read  from  STDIN.  This  can  only  be  used  if  only  one
          passphrase  is  supplied.  Note that this passphrase is only used if
          the option --batch has also been given.  This is different from gpg.

   --passphrase-file file
          Read the passphrase from file file. Only the first line will be read
          from file file. This can only be used if only one passphrase is sup‐
          plied. Obviously, a passphrase stored in a file is  of  questionable
          security if other users can read this file. Don't use this option if
          you can avoid it.  Note that this passphrase is  only  used  if  the
          option --batch has also been given.  This is different from gpg.

   --passphrase string
          Use  string  as  the  passphrase.  This can only be used if only one
          passphrase is supplied. Obviously,  this  is  of  very  questionable
          security  on  a  multi-user system. Don't use this option if you can
          avoid it.  Note that this passphrase is  only  used  if  the  option
          --batch has also been given.  This is different from gpg.
```
The first is being used, the third would be the easiest. But without verifying if the password would appear in `ps` or whatnot it shouldn't be used :D 

Another way would be to change the file descriptor for `$password`
The resulting function could look like this (**untested**):
```shell
gpg_decrypt() {
        # gpgver could be removed if 2.0 if of no concern?
	local gpgver=$(gpg --version --no-permission-warning | awk '/^gpg/ {print $3}')
	local gpgpass="$1"
	local tmpres ret
	typeset -a gpgopt
	gpgpopt=(--batch --no-tty --passphrase-fd 3 --no-options)

	{ option_is_set -g } && {
		gpgpass="$TOMBKEY"
		gpgpopt=(--yes)

		# GPG option '--try-secret-key' exist since GPG 2.1
		{ option_is_set -R } && { autoload -U is-at-least && is-at-least "2.1" $gpgver } && {
			typeset -a recipients
			recipients=(${(s:,:)$(option_value -R)})
			{ is_valid_recipients $recipients } || {
				_failure "You set an invalid GPG ID."
			}
			gpgpopt+=(`_recipients_arg "--try-secret-key" $recipients`)
		}
	}

	_tmp_create
	tmpres=$TOMBTMP
	TOMBSECRET=`print - "$TOMBKEY" | \
		gpg --decrypt ${gpgpopt[@]} \
			--status-fd 2 --no-mdc-warning --no-permission-warning \
			--no-secmem-warning 3<<<"$gpgpass" 2> $tmpres`
	unset gpgpass
	ret=1
	for i in ${(f)"$(cat $tmpres)"}; do
		_verbose "$i"
		[[ "$i" =~ "DECRYPTION_OKAY" ]] && ret=0;
	done
	return $ret
}
```
Alternatively a anonymous pipe could be utilized in combination with `--passphrase-file` (**untested**):
```shell
gpg_decrypt() {
	# again only needed if 2.0 is of concern
	local gpgver=$(gpg --version --no-permission-warning | awk '/^gpg/ {print $3}')
	local gpgpass="$1"
	local tmpres ret
	typeset -a gpgopt
	gpgpopt=(--batch --no-tty --passphrase-file <(echo "$gpgpass") --no-options)

	{ option_is_set -g } && {
		gpgpass="$TOMBKEY"
		gpgpopt=(--yes)

		# GPG option '--try-secret-key' exist since GPG 2.1
		{ option_is_set -R } && { autoload -U is-at-least && is-at-least "2.1" $gpgver } && {
			typeset -a recipients
			recipients=(${(s:,:)$(option_value -R)})
			{ is_valid_recipients $recipients } || {
				_failure "You set an invalid GPG ID."
			}
			gpgpopt+=(`_recipients_arg "--try-secret-key" $recipients`)
		}
	}

	_tmp_create
	tmpres=$TOMBTMP
	TOMBSECRET=`print - "$TOMBKEY" | \
		gpg --decrypt ${gpgpopt[@]} \
			--status-fd 2 --no-mdc-warning --no-permission-warning \
			--no-secmem-warning 3<<<"$gpgpass" 2> $tmpres`
	unset gpgpass
	ret=1
	for i in ${(f)"$(cat $tmpres)"}; do
		_verbose "$i"
		[[ "$i" =~ "DECRYPTION_OKAY" ]] && ret=0;
	done
	return $ret
}
```
This will make the password available via a file descriptor (as an anonymous pipe normally as /dev/fd/63). The lifetime is bound to the runtime of `gpg`. And only gpg should be able to read or write from this location. And regarding the visibility: It should only show `/dev/fd/63` in the process monitor tools.

Light testing via those commands:
```shell
$ zmodload zsh/mapfile
$ FILECONT=${mapfile[test.key]}
$ print - "$FILECONT" | gpg --decrypt --batch --no-tty --no-options --status-fd 0 --passphrase-file <(echo "test")
gpg: AES256.CFB verschlüsselte Daten
gpg: Verschlüsselt mit einem Passwort
(...)
$ print - "$FILECONT" | gpg --decrypt --batch --no-tty --no-options --status-fd 0 --passphrase-fd 3 3<<<"test"
gpg: AES256.CFB verschlüsselte Daten
gpg: Verschlüsselt mit einem Passwort
(...)                                                      
```

But before looking further that, I would like to get rid of the restriction `gpg1` :D

Narrat:
As a draft  because of the same reasons mentioned in #585 
___________________
Background of this PR is the issue that tomb has issues with passwords that try to use characters in a combination so they create control sequences for the shell. For example newline `\n`, form feed `\f` and tabulator `\t`.
This is due how the password is provided for the gpg call. gpg is used with `--passphrase-fd 0` which reads the password from `STDIN` until it encounters a `newline`. Unfortunately this means the shell interprets those character sequences in the password itself. Especially fatal in case of \n as this will reduce the password to this point. Example: you want to set password `test\ntest`. gpg, while reading `STDIN`, will stop at `\n` and the resulting password will be only `test`.
Not sure yet what happens with the rest, but it does seem to be discarded in general and not added to or used as `TOMBSECRET`.

Two ways to avoid interpreting control sequences:

* change `--passphrase-fd` to a dedicated descriptor above 2 (like `--passphrase-fd 3` and input password like `3<<<"$password"`
 * use `--passphrase-file` and use an anonymous pipe

In general the first option is similar to the current solution, just changing it to a different FD, which allows to avoid interpreting control sequences. But is untested if the redirect isn't visible while tracing.
This solutions changes the fd for the password input from STDIN to a new one file-descriptor (3 in this example).

### DIFF
--- a/ChangeLog.md
+++ b/ChangeLog.md
@@ -1,4 +1,20 @@
 # Tomb ChangeLog
+
+## Unreleased
+### TBA
+
+Special character sequences like newline (`\n`), tabfeed (`\t`) and
+similar from ANSI Escape Sequences caused unexpected behaviour when
+setting passwords. The safe way to pipe the passwords into `gnupg`
+would go through the shell and if sequences were encountered
+they would be processed.
+Which brings unintended changes in the entered password. In the case
+of a newline the user could end up with a much weaker password
+than expected and not notice it necessarily.  
+With this release the password won't be changed by the shell pipe
+anymore. `Pinentry` and `libassuan` still do their thing though.  
+For an overview see [Known Bugs](./KNOWN_BUGS.md).
+
 ## 2.13
 ### July 2025
 
--- a/KNOWN_BUGS.md
+++ b/KNOWN_BUGS.md
@@ -1,3 +1,41 @@
+# Passwords getting silently altered
+## Issues with Tomb version x.x to 2.13
+
+The safe way of piping a password into gnupg would be interpret by the shell
+and therefore character sequences like `ANSI Escape Sequences` could cause
+unintended changes to the password and not necessarily be noticeable.
+But also escaping characters can introduce changes to the final password.
+In general a tomb created with 2.13 and before can still be unlocked, but
+it needs to happen with unsafe options.  
+Below is a table with some examples:
+
+| Sequence | Theoretical | Real | Unlock 2.13 | Unlock >2.13 | Notes |
+|----------|-------------|------|-------------|--------------|-------|
+| newline  | `test\ntest` | `test` | `test` or `test\ntest` or `test\nwhatever` | `test` | Position matters greatly
+| horizontal tab  | `test\test` | `test<TAB>est` | `test\test` | `--tomb-pwd "test\test"` |
+| formfeed | `test\ftest` | `test\xCtest` | `test\ftest` | `--tomb-pwd "test\ftest"` | dm-crypt will throw a warning |
+| bell  | `test\atest` | `test\x07test` | `test\atest` | `--tomb-pwd "test\atest"` | bell will appear in output
+| backspace  | `test\btest` | theoretically `testest`? | `test\btest` | `--tomb-pwd "test\btest"` | behaves unexpectedly
+| vertical tab  | `test\vtest` | `test<VTAB>test` | `test\vtest` | `--tomb-pwd "test\vtest"` |
+| carriage return  | `test\rtest` | `test<CR>test` | `test\rtest` | `--tomb-pwd "test\rtest"` |
+| escape character (wrong)  | `test\etest` | in shell `testest` | `test\etest` | `--tomb-pwd "test\etest"` | behaves unexpectedly
+| escape character | `test\e[1Ktest` | in shell `    test` | `test\e[1Ktest` | `--tomb-pwd "test\e[1Ktest"` | erase from start to cursor
+| escpape backslash | `testtest\` | - | - | - | created key is safely protected...
+
+It isn't explored what happened in cases like `\ntest` or `test\`.
+In the first case would `test` be discarded or end up as the `LUKS` key?
+Essentially the same question for the second case:
+What does the `LUKS` key look like after the password was appended with the
+intended key?  
+The general advice is to generate a new tomb to be on the safe side.
+
+Of note: This doesn't address the quirk with `%`s in a password.
+Pinentry or rather Assuan have the requirement, that some special sequences
+need to be percent escaped (`%` as `%25`, `CR` as `%0D` and `LF` as `%0A`).
+If such a password was set via pinentry and is later supplied via
+`--unsafe --tomb-pwd` it will need to be set as `--tomb-pwd test%25test`.
+
+
 # Password bug failing to open tombs
 ## Issue with Tomb version 2.12 (short lived)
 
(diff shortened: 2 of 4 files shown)


jaromil:
I'm in favor of this, yet we'll need to state the introduction of a breaking change for anyone who uses passwords with control characters which were previously misinterpreted.

